PR #206, opened 2015-10-12
Converted StarterWeb's 'min:js' and 'min:css' Gulp tasks to asynchronous

scottaddie:
As a best practice, Gulp tasks should return a stream. This has the added benefit of making such tasks asynchronous. See the official Gulp docs here: https://github.com/gulpjs/gulp/blob/master/docs/API.md#return-a-stream.


pranavkm:
cc @DamianEdwards

spboyer:
:+1:

rustd:
@madskristensen FYI. Does this look good? It looks good to me.

rustd:
@phenning I chatted with @madskristensen and we are good to merge this in.

PR #209, opened 2015-10-12
React to Microsoft.Framework -> Microsoft.Extensions renames

pranavkm:


pranavkm:
cc @phenning

PradeepKadubandi:
This particular change looks good to me!

I have a couple of questions in general for template crew (@phenning - perhaps you can work with Pranav Rastogi, not sure if these were already considered though)
- Are we including any unneeded namespaces ?(If I choose Remove and Sort usings, will there be any changes in files?) If there are some unused ones, is that intentional?
  - Second and somewhat individual preference, so the usings should be within the namespace declaration?

pranavkm:
Historically usings have always been located outside the namespace declaration so we might want to stick with that layout for templates. Not sure if we have extra namespaces - perhaps file a bug for it?

Issue #125, opened 2015-07-08
Update jquery-validation-unobtrusive to 3.2.3


danroth27:

rustd:
Done for RC1

PR #212, opened 2015-10-13
Address #125 update bower dependencies

spboyer:
- Update jquery-validation-unobtrusive to 3.2.3 #125
- change to bower.json
- update `bower_components` files


spboyer:
Would suggest going with semver for bower and npm dependencies.

_Tilde ~_
For tilde ranges, major and minor versions must match those specified, but any patch version greater than or equal to the one specified is valid.

For example, ~1.2.3 permits versions from 1.2.3 up to, but not including, the next minor, 1.3.0.

_Caret ^_
For caret ranges, only major version must match. Any minor or patch version greater than or equal to the minimum is valid.

For example, a range of ~1.2.3 will only permit versions up to, but not including 1.3.0. However, the caret version, ^1.2.3 permits versions from 1.2.3 all the way up to, but not including, the next major version, 2.0.0.

Would also suggest __not__ packaging or including the npm and bower resources in the templates as the generator-aspnet are not and using the package tools to install these resources.

@peterblazejewicz @sayedihashimi

peterblazejewicz:
It appears that project will be community driven (I can see beta5 commits everywhere), so I'd use default one to most used one (seems to be NPM). The Node/NPM now uses caret as default save prefix and the only way to change this behaviour is to use local settings (`npm config set save-prefix '~'`). The one prefixer is easier to maintain and it is up to community to update everywhere, similar:
https://github.com/yeoman/generator-angular/blob/master/templates/common/root/_bower.json

There should be separate issue started how do they plan to handle dependencies - I think they should be not stored in repository - but only released with templates via tools - e.g. via VS. People using this repository directly should be aware of using tools like NPM/Bower - just presumption but could be correct.

rustd:
VS templates need an offline install so the components are laid out as it is. This is a separate issue.
@phenning This PR is good to merge. However you would have to generate the vs template again as the files have changed.

rustd:
@phenning we will have to update it to 3.2.4 as per this https://github.com/aspnet/Templates/issues/239

spboyer:
@rustd updated this PR to address #239 cc:/ @phenning

rustd:
thanks! @phenning should be good to be merged.

phenning:
We don't have the bower_components any more, any content there should go into wwwroot\lib. (see .bowerrc)

peterblazejewicz:
> We don't have the bower_components any more, any content there should go into wwwroot\lib. (see .bowerrc)

@phenning 
There is no source for `.bowerrc` on this project

spboyer:
@phenning I'll add the .bowerrc and remove the bower_components dir. This will put the structure inline with what you want.

phenning:
.bowerrc is here:

https://github.com/aspnet/Templates/blob/dev/src/BaseTemplates/StarterWeb/.bowerrc

peterblazejewicz:
@phenning 
Apologies, somehow ES in GitHub ignores them: https://github.com/aspnet/Templates/search?utf8=%E2%9C%93&q=.bowerrc

phenning:
@peterblazejewicz no problem, still weird it doesn't show up on search.

spboyer:
I'll fix the location no problem, need about an hr. On the road

peterblazejewicz:
> @peterblazejewicz no problem, still weird it doesn't show up on search.

Solved, FYI: 'bowerrc in:path'  https://help.github.com/articles/searching-code/#scope-the-search-fields

spboyer:
@phenning ok, this is set now. `bower_components` is now gone and `wwwroot\lib` has been reset.

yukozh:
:watch:

phenning:
Merged 19294a2565f1542e68a02cf27a495f88fc30ac94 and 2cd948ef18d44fbd8ef4bad8708a035e1a13d674 after doing some creative squashing so we wouldn't bring bower_components back into the history.

PR #215, opened 2015-10-13
:bug: Use .sr-only class for fallbacks. Closes #150

### LINKED ISSUE #150: Bootstrap: Update the check for fallback logic
The fallback logic does not work with bootstrap v3.3 since they changed the value. We need to update it in the template so it works for all version of bootstrap

peterblazejewicz:
The current fallback implementation for Bootstrap 3.\* fails when using
recent version of Bootstrap as the declaration for .hidden class has
changed. The .sr-only class was introduced with 3.0.0 release and is still
there in 3.3.5:
http://git.io/vC26P
#150

The change is future-proof for Bootstrap 3.3.5 in case there is update to BS version in Templates.

Thanks!


rustd:
:shipit: @phenning this should be good to merge.

PR #219, opened 2015-10-15
React to Auth changes

HaoK:
cc @Tratcher (appears OpenID templates were probably broken since the switch from IOptions => instance since they were using ConfigureServices)


rustd:
:shipit:

phenning:
:shipit:

HaoK:
You guys want me to check this in now?  Its still green on CI (but it shouldn't be :)

phenning:
Go ahead and merge it, the CI doesn't run org auth tests (other then just compiling that is). We do the org auth testing over in the tooling branch after we integrate the templates.

HaoK:
329552da82bb80ef1d26378185cc5e7ba7ecb45b

PR #227, opened 2015-10-18
Removed some old code and fix the removelogin bug

Anderman:


rustd:
This looks good. @Anderman why are you changing the RemoveLogin.cshtml file. It should not require any change.

Anderman:
I think this is replaced by the manageLogin.chtml. I could not find any reference to it. So I removed this file

rustd:
Thanks. yes that is the case. :shipit: @phenning This should be good to be merged

phenning:
Can you remove the corresponding lines that reference the removed file from Templates.xml andTemplateRules.csproj? The rest of the change looks good.

https://github.com/aspnet/Templates/blob/dev/src/Templates.xml
https://github.com/aspnet/Templates/blob/dev/src/TemplateRules.csproj

Anderman:
Yes, I forgot that, I opened this project in VS2015, But VS will make a mess about the layout. That's why I didn't commit this file

phenning:
I just noticed two other issues while getting the merge ready. Can you please _add_ the RemoveLoginViewModel.cs to the Template.xml and TemplateRules.csproj? Also, I made an inline comment as well in regards to coding-styling.

Anderman:
Almost I also have to add the view model to the project and xml file

Anderman:
In The last commit  I removed an 'x' . I think it was a typo on this commit https://github.com/aspnet/Templates/commit/8c8b846470216b18bb37eb2901bd49105aa0fc58

phenning:
Looks good! Thanks for removing the typo, the data model the tooling uses is resilient to it though! I'll merge it now.

phenning:
Squashed some of the commits and merged.

PR #233, opened 2015-10-19
:bug: Use proper names in fallback tests. Fixes #232

### LINKED ISSUE #232: Validation script fallbacks fail due to all lower case names for `jquery` global object
The code for validation scripts fallbacks in validation scripts partial:
https://github.com/aspnet/Templates/blob/5605c55164a872fcf839d24cbcb1caa5b7e9306c/src/Rules/StarterWeb/IndividualAuth/Views/Shared/_ValidationScriptsPartial.cshtml
uses all lower case script, so it fails, which is unfortunate. The `jquery` as name should be `jQuery`
Thanks!

peterblazejewicz:
- use `jQuery` instead of `jquery` as object name

This commit fixes failing fallback tests for validation
scripts that are failing at runtime due to incorrect name of jQuery object
used in evaluated script. The commit introduces the same notation as used in the
main scripts fallbacks.

Fixes #232


rustd:
:shipit: @phenning this is good to be merged.

PR #243, opened 2015-10-21
Update HomeController.cs

yukozh:


rustd:
:shipit: @phenning this can be merged.

phenning:
if you are going to update it here, can you also update it here?

Rules\StarterWeb\OrganizationalAuth\Common\Controllers\HomeController.cs

peterblazejewicz:
Q: Which view will be rendered when Error middleware fires /Home/Error action?

yukozh:
@peterblazejewicz This is the razor view engine problem. it is able to find error.cshtml automaticly. you can see the source in mvc repo. /Home/Error is related on routing, not view engine. Here, it is just a rule for searching view. We can also use default rule to find it. So, I created this PR.

yukozh:
thanks a lot.

peterblazejewicz:
@Kagamine Thanks!

PR #244, opened 2015-10-21
React to API change in UseDatabaseErrorPage

rowanmiller:
The overload being used in the templates has swapped to a builder pattern (see https://github.com/aspnet/Diagnostics/issues/184). But rather than updating to the new signature, just swapping to the parameterless overload since it enables everything by default (same as UseDeveloperExceptionPage()).


phenning:
Is this due to an API change and is dependent on a newer coherence?

rowanmiller:
It is due to an API change, but its not dependent on a new coherence since the overload I am swapping to is present in the current builds.

rowanmiller:
Updated with one more instance of the method call that I missed in the initial changes

phenning:
Awesome, :shipit:

PR #246, opened 2015-10-21
Add text context to toggle buttons

peterblazejewicz:
This commit adds textual context to _hamburger_ toggle menu
buttons used in navigation. The solution uses built-in Bootstrap
helper class that hides content from page but not from screen readers
and crawlers - .sr-only. The change is future proof as BS 4 alpha seems
to use the same helper class

The markup uses solution documented and used in default Bootstrap 3.\* templates:
http://getbootstrap.com/components/#navbar-default

After the change screen readers are able to see text context of toggle button:

image

Just a detail.

Thanks!

### DIFF
--- a/src/BaseTemplates/StarterWeb/Views/Shared/_Layout.cshtml
+++ b/src/BaseTemplates/StarterWeb/Views/Shared/_Layout.cshtml
@@ -25,6 +25,7 @@
         <div class="container">
             <div class="navbar-header">
                 <button type="button" class="navbar-toggle" data-toggle="collapse" data-target=".navbar-collapse">
+                    <span class="sr-only">Toggle navigation</span>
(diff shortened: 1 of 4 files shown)


rustd:
Thanks! @peterblazejewicz Instead of just fixing this, I think it would be good to do a pass on all pages and see if there are any other that we should update.

peterblazejewicz:
Yes, but the PR will diverge. This change is single, contained and focused on navbar toggle
Up to you though :)

phenning:
@peterblazejewicz Can you open a separate issue to investigate whether there is anywhere else we should be using .sr-only?  We'll take this PR as is for this specific instance.

peterblazejewicz:
@phenning Done #249

phenning:
Thanks!

PR #248, opened 2015-10-22
Use IsSignedIn() instead of existing api in _LoginPartial.cshtml

### LINKED ISSUE #56: Use IsSignedIn() instead of existing api in _LoginPartial.cshtml
Had a chat with @HaoK  and in the existing _LoginPartial.cshtml file the way we check if user is logged in is using `User.Identity.Authenticated`. It should be updated to use `User.IsSignedIn()`

phenning:
Fixes #56 

Also updated the controllers. Individual auth controller was already using the proper API.

@HaoK @rustd

### DIFF
--- a/src/Rules/StarterWeb/IndividualAuth/Views/Shared/_LoginPartial.cshtml
+++ b/src/Rules/StarterWeb/IndividualAuth/Views/Shared/_LoginPartial.cshtml
@@ -1,6 +1,6 @@
 ﻿@using System.Security.Claims
 
-@if (User.Identity.IsAuthenticated)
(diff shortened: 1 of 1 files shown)


rustd:
let's make this change only for Individual Auth template. Not sure if this works with Org Auth. @HaoK do you know?

HaoK:
I would try the changes before you check them in yes :)

phenning:
Updated the PR, only doing individual auth now, we're not pulling in the package that has the extension in the template, so minimizing this change for now.

phenning:
Merged to dev

PR #252, opened 2015-10-23
fix #251: Remove SetBasePath(appEnv.ApplicationBasePath) from Templates

### LINKED ISSUE #251: Remove SetBasePath(appEnv.ApplicationBasePath) from Templates
This is inferred now and doesn't need to be set.

/cc @divega @kirthik

kirthik:
cc @phenning


phenning:
:shipit:

PR #256, opened 2015-10-23
Fix relogin issue in organization auth

### LINKED ISSUE #255: Organizational account authentication broken in beta8? Forces relogin on every request.
_From @unt1tled on October 23, 2015 1:50_

I'm able to File-New-Project a new Web Application with Organizational Account authentication and successfully login. But ever since beta8, I'm unable to navigate to any page other than Home/Index. It seems as though each [Authorize]d request redirects to the OpenId login page, which in turn redirects me back to the Index Page.

_Copied from original issue: aspnet/Security#544_

phenning:
Fixes #255 

@Tratcher, @rustd


Tratcher:
@HaoK

HaoK:
AutomaticAuthentication is going to become AutomaticAuthenticate once you pick up new packages.  Otherwise :shipit:

Issue #257, opened 2015-10-23
Move webroot:wwwroot from project.json to Microsoft.AspNet.Hosting.json


Tratcher:
https://github.com/aspnet/Templates/blob/dev/src/BaseTemplates/EmptyWeb/project.json#L2
Microsoft.AspNet.Hosting.json:

```
{
  "Hosting:WebRoot": "wwwroot"
}
```

Or

```
{
  "Hosting":
  {
     "WebRoot": "wwwroot"
  }
}
```

@DamianEdwards? I don't recommend promoting WebRoot to a top-level element.
https://github.com/aspnet/dnx/issues/3035
https://github.com/aspnet/Hosting/issues/411

Tratcher:
@davidfowl

rustd:
@Tratcher rc1?

Tratcher:
Yes, rc1

phenning:
This has some tooling impact as well.

Tratcher:
Feedback: remove the Microsoft.AspNet.Hosting.json file from the default templates. The runtime uses intelligent defaults and the file is only needed for customization.

scottaddie:
The `gulpfile.js` file included in the StarterWeb template will need to be adjusted accordingly. It currently pulls in the `project.json` file via `require` and then accesses its `webroot` property.

rustd:
@Tratcher Let's remove the file if the runtime has defaults. The gulpfile will have to be updated to hard code wwwroot as a path.
@Tratcher are you reverting your change?

Tratcher:
I'll take care of the updates today

PR #262, opened 2015-10-26
Move webroot from project.json to Microsoft.AspNet.Hosting.json.

Tratcher:
#257 @davidfowl @BrennanConroy @rustd @phenning

### DIFF
--- a/src/BaseTemplates/EmptyWeb/Microsoft.AspNet.Hosting.json
+++ b/src/BaseTemplates/EmptyWeb/Microsoft.AspNet.Hosting.json
@@ -1,0 +1,3 @@
+{
+  "webroot": "wwwroot"
+}
--- a/src/BaseTemplates/EmptyWeb.vstemplate
+++ b/src/BaseTemplates/EmptyWeb.vstemplate
@@ -18,6 +18,7 @@
   </TemplateData>
   <TemplateContent>
     <Project File="EmptyWeb.xproj" ReplaceParameters="true">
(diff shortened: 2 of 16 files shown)


src/BaseTemplates/EmptyWeb/Microsoft.AspNet.Hosting.json
```diff
@@ -0,0 +1,3 @@
+{
+  "Hosting:WebRoot": "wwwroot"
```
davidfowl: Change the casing to be "hosting:webRoot":

Tratcher: how about just "hosting:webroot"?

davidfowl: better

Tratcher:
Updated

phenning:
Also needs to be removed in the project.json files under src\Rules.

Tratcher:
Updated @phenning

Tratcher:
Or @phenning, do I not need to add Microsoft.AspNet.Hosting.json to all of the Rules projects because it's not different from the base project?

davidfowl:
:shipit: We should just put the server name in here too :smile:

phenning:
@Tratcher You only need to add it to the base templates if the content is identical between ind auth/no auth, etc.

Tratcher:
@davidfowl That would make it harder to have multiple server commands.

Issue #188, opened 2015-09-30
Add TargetFrameworkVersion to .xproj files


phenning:
We may need this in order to properly handle frameworks other then 4.5.1

/cc @BillHiebert

phenning:
After talking to @BillHiebert, we decided to punt this to RC2 for more bake-time since there is currently nothing explicitly broken at the moment.

PR #266, opened 2015-10-27
Add TargetFrameworkVersion to xproj files.

phenning:
#188

/cc @BillHiebert @davidfowl @rustd


davidfowl:
How is this correct? Why wouldn't it be in the targets file instead of the template?

BillHiebert:
It's not, and the targets file is worse since it is shared across all projects. At least this allows the projects created as 4.5.2, 4.6 etc, to advertise their target framework correctly.  I am not sure what the right thing to do is since it should reflect what is in project.json and never be hardcoded like this. Maybe we should be setting that property via a target/task in msbuild which reads the value from the json file. But even that is problematic if there is more than one framework defined in project.json.

phenning:
We're going to hold off on this PR until post RC1, since we don't see anything obvious that is broken at the moment.

PR #271, opened 2015-10-28
Don't add Microsoft.AspNet.Hosting.json by default. 

Tratcher:
#257 @davidfowl @rustd @phenning

And hard code wwwroot in gulpfile.js.


davidfowl:
:shipit:

phenning:
:shipit:

PR #276, opened 2015-10-29
Define gulpfile.js asset paths in more terse manner

scottaddie:
The file paths defined in the gulpfile.js file can be defined in a more concise fashion, so that `paths.` doesn't have to be keyed over and over again when adding new properties to the `paths` object literal.

### DIFF
--- a/src/BaseTemplates/StarterWeb/gulpfile.js
+++ b/src/BaseTemplates/StarterWeb/gulpfile.js
@@ -7,16 +7,16 @@ var gulp = require("gulp"),
     cssmin = require("gulp-cssmin"),
     uglify = require("gulp-uglify");
 
-var paths = {
-    webroot: "./wwwroot/"
-};
+var webroot = "./wwwroot/";
 
-paths.js = paths.webroot + "js/**/*.js";
-paths.minJs = paths.webroot + "js/**/*.min.js";
-paths.css = paths.webroot + "css/**/*.css";
-paths.minCss = paths.webroot + "css/**/*.min.css";
-paths.concatJsDest = paths.webroot + "js/site.min.js";
-paths.concatCssDest = paths.webroot + "css/site.min.css";
+var paths = {
+    js: webroot + "js/**/*.js",
+    minJs: webroot + "js/**/*.min.js",
+    css: webroot + "css/**/*.css",
+    minCss: webroot + "css/**/*.min.css",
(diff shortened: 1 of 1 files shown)


rustd:
This looks good to me. @madskristensen ?

madskristensen:
Looks good. Shouldn't we get the `webroot` value from a settings file?

scottaddie:
@madskristensen Agreed. What about a custom entry in `package.json`? It might look as follows:

``` javascript
"-vs-webroot": "wwwroot"
```

We would just have to `require` the `package.json` file in `gulpfile.js` again. That was recently removed. Unless you have a recommended convention to follow, I suggest using the "-vs" prefix because that's what your "NPM Scripts Task Runner" extension uses for script bindings. Thoughts?

madskristensen:
The webroot will be in some settings file, but I don't know which one. We will not use package.json for that. @DamianEdwards do you know?

DamianEdwards:
It will be in `hosting.json`

scottaddie:
@DamianEdwards @madskristensen I thought the decision was made to remove `hosting.json` from the StarterWeb template (see https://github.com/aspnet/Templates/issues/257). Are you okay with me adding it back to store the web root path?

madskristensen:
You're thinking of hosting.ini. This is different.

scottaddie:
@madskristensen I do recall that one being removed, but there was a `Microsoft.AspNet.Hosting.json` file. It was in the StarterWeb template repo for a short while, there was talk about renaming it to `hosting.json`, and then it was quickly removed.

@Tratcher Do you know anything about this? https://github.com/aspnet/Templates/issues/257

rustd:
The hosting file was removed since the runtime default webroot to wwwroot. We should not be adding the file back.

madskristensen:
Oh, then hardcoding the webroot into gulpfile.js is fine

scottaddie:
@madskristensen Sounds good. Is this one good to merge now?

madskristensen:
Yes, looks good to me

rustd:
@scottaddie Can you please make these changes in dev

scottaddie:
@rustd Sorry, but I'm confused. I thought the process was to create a branch in my fork and submit a PR with that so that it could be merged into the dev branch. What am I missing?

rustd:
@scottaddie No you are good. @joeloff this can be merged into dev.

PR #285, opened 2015-10-30
Default configuration for WebAPI template

logic01:
Be kind this is my first time using git.


rustd:
@logic01 what issue are you fixing? As a best practice, please open an issue first before sending the PR so we can look at what the problem is.

logic01:
Closed so I can create an Issue and make the appsettings.json change recomended by rustd and ashimoon.

Issue #287, opened 2015-10-30
Add default configuration to WebApi template.


logic01:
Currently the WebApi template does not contain any default configuration file or initialization there of.

rustd:
@logic01  As part of this https://github.com/aspnet/Templates/issues/286 we update the Web API template to include Configuration. 
Thank you for reporting this issue.

Issue #286, opened 2015-10-30
Add sensible defaults to templates that use logging


rynowak:
We've added a bunch of features to logging to improve the local dev story for use with the console logger. Most importantly we want to provide a sensible set of default logging levels for the console logging experience so that users aren't overwhelmed with spam. We want this filtering to more restrictive around log messages provided by the system, and not restrictive around log messages added by the developer, so a single level won't do.

This change adds a configuration file for logging as well as the startup code to wire it up. The configuration file approach is important as a jumping off point for more fine-tuning of what messages are included and excluded.

**Change to constructor**

```
            Configuration = builder.Build();
```

**Becomes**

```
            Configuration = builder.Build().ReloadOnChanged("appsettings.json");
```

**Change to `Configure(...)`**

```
            loggerFactory.MinimumLevel = LogLevel.Information;
            loggerFactory.AddConsole();
            loggerFactory.AddDebug();
```

**Becomes**

```
            loggerFactory.AddConsole(Configuration.GetSection("Logging"));
            loggerFactory.AddDebug();
```

**Add to appsettings.json**

```
{
  "Logging": {
    "IncludeScopes": false,
    "Switches": {
      "Default": "Verbose",
      "System": "Information",
      "Microsoft": "Information"
    }
}
```

**Add to project.json**

```
  "dependencies": {
    ...
    "Microsoft.Extensions.Configuration.FileProviderExtensions" : "1.0.0-*",
    ...
  },
```

rustd:
@phenning we just updated the code changes.

phenning:
e6d34253635d243e1d639d51b98ec038c00f1998

PR #288, opened 2015-10-30
Fix issue #287 - Default configuration for WebApi template

logic01:
Fixed to fit beta 8 standard.


rustd:
@logic01 Thank you for the PR. This was fixed when we fixed https://github.com/aspnet/Templates/issues/286

PR #303, opened 2015-11-05
Align `build.cmd` with latest Universe templates

dougbu:
- use latest `NuGet.exe`
- use specified (often prerelease) KoreBuild package
- avoid `dnvm upgrade` because it messes with the user-level `$env:Path` setting

nit: remove trailing whitespace

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -1,8 +1,11 @@
-@echo off 
+@echo off
 cd %~dp0
 
 SETLOCAL
-SET CACHED_NUGET=%LocalAppData%\NuGet\NuGet.exe
+SET NUGET_VERSION=latest
+SET CACHED_NUGET=%LocalAppData%\NuGet\nuget.%NUGET_VERSION%.exe
+SET BUILDCMD_KOREBUILD_VERSION=
+SET BUILDCMD_DNX_VERSION=
 
 IF "%*"=="" (
   SET SCRIPTARGS=verify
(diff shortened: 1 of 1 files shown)


dougbu:
/cc @phenning 

FYI build buddies: This PR should prevent future unintended changes to the asplab `$env:Path` setting.

build.cmd
```diff
@@ -10,25 +13,34 @@ IF "%*"=="" (
...
-@powershell -NoProfile -ExecutionPolicy unrestricted -Command "$ProgressPreference = 'SilentlyContinue'; Invoke-WebRequest 'https://www.nuget.org/nuget.exe' -OutFile '%CACHED_NUGET%'" 
- 
-:copynuget 
-IF EXIST .nuget\nuget.exe goto restore 
-md .nuget 
-copy %CACHED_NUGET% .nuget\nuget.exe > nul 
- 
-:restore 
-IF EXIST packages\KoreBuild goto run  
-.nuget\NuGet.exe install KoreBuild -ExcludeVersion -o packages -nocache -pre -Source https://myget.org/F/aspnetvnext
```
dougbu, line 25: Not sure about removing the `-Source` argument here. The change won't affect CI builds because earlier steps create a `NuGet.config` file containing the Right Stuff:tm:.

@joeloff any expected issues for local builds i.e. to most people have reasonable default feeds, containing `https://myget.org/F/aspnetvnext` or newer? I'll add a `NuGet.config` file if you think it's worthwhile.

joeloff, line 25: For the most part the template build pulls in packages from the internal machines and use those as source. I'd assume most people do have good feeds. If it trips people up we can probably look at adding a config file

dougbu:
Is this ready @joeloff ?

PR #305, opened 2015-11-08
remove unnecessary usages of await

TerribleDev:
awaiting a return of a task object is unnecessary


davidfowl:
:shipit:

rustd:
@joeloff can you please merge this into dev

PR #319, opened 2015-11-15
Remove duplicate declarations for .carousel-caption class

peterblazejewicz:
The z-index property is already declared in original Bootstrap CSS:

``` css
.carousel-caption {
  position: absolute;
  ...
  bottom: 20px;
  z-index: 10;
  padding-top: 20px;
  ...
```

http://git.io/v4GRF
The duplicated declarations are only adding noise to source code
of templates - which are supposed to be used as guides (why someone
should start to wonder what is a purpose of these z-index declarations?)

Thanks!


rustd:
:shipit: @joeloff this should be merged into dev

Issue #316, opened 2015-11-13
Use C# expression-bodied members to cleanup MVC controller code


scottaddie:
In the "StarterWeb" template's `HomeController.cs` file, there's not much to the `Index` and `Error` action methods. This would be a perfect opportunity to demonstrate the use of C# expression-bodied members. I see the benefits as two-fold:
1. It would raise awareness of a new language feature
2. It would cleanup the code a little bit

Here are the exact changes I'm proposing:

``` c#
public IActionResult Index() => View();
public IActionResult Error() => View();
```

If everyone is okay with this idea, I'll submit a PR for it.

herecydev:
:+1:

Eilon:
I would say it's a non-goal to use the project templates to raise awareness of new language features. Otherwise we'd fill up the templates with lots of new syntax to demo the features.

So the question really is whether the new code is cleaner or not and still is useful for the sake of the template.

One thing to keep in mind with the templates is that they sometimes have extra stuff in them because the idea is that the template is not already a finished application. Rather, it is intended as a starting point that has a lot of details left to be filled in. Do we expect more code to be added to the `Index()` and `Error()` actions? I think for the `Index()` action the answer is definitely _yes_. So if it's changed to use an expression-bodied member, it becomes _more_ work for the user to convert it back to a method-bodied method (I just invented a new term :smile: ) so that the code can be added.

scottaddie:
@Eilon Knowing that feature awareness isn't a goal of the templates, I do see your point with the `Index` action method. It likely will be modified by the developer. However, I would still argue that the default `Error` action method sees much more infrequent modification. What about applying the new syntax just to that method?

Eilon:
For the `Error` method I could see it either way, no strong opinion from me. The new templates (as of RC1) do already have an expression-bodied member in them: https://github.com/aspnet/Templates/blob/dev/src/Rules/StarterWeb/AI/NoAuth/Startup.cs#L74

So, flip a coin perhaps :smile:

rustd:
I don't think we should change the Error method. If the user wants to modify the method, this pattern makes it harder. This pattern makes it harder to customize the code. In the Startup.cs chances are they will not modify it so it is fine to have it but IMO the Error method is something that should be easy to customize.

rustd:
@scottaddie I am closing this issue as a Won't Fix.

PR #322, opened 2015-11-18
Use expression bodied member in HomeController Error method

scottaddie:
I modified the StarterWeb project's HomeController Error action method to use a C# expression bodied member. My experience has been that this action method rarely deviates from the default template code, making it a perfect fit for this new C# feature. I considered doing the same for the Index action method; however, that method does often consist of much more than a return statement.

See [this issue](https://github.com/aspnet/Templates/issues/316) for the discussion leading up to this PR.


phenning:
Closing this since the underlying issue was closed as Won't Fix.

Issue #324, opened 2015-11-19
Remove duplicate btn-default class


sayedihashimi:
In https://github.com/aspnet/Templates/blob/dev/src/BaseTemplates/StarterWeb/Views/Home/Index.cshtml we have duplicate instances of `btn-default`.

If you recall one of the first PRs here https://github.com/aspnet/Templates/pull/210 was from @spboyer to fix this here.

As you know we get content from this repo and push to [generator-aspnet](https://github.com/OmniSharp/generator-aspnet/).

We've already fixed this two times in generator-aspnet. I don't understand why this keeps coming back, it's really annoying and an indication that the content here is not being carefully reviewed.

@barrytang @phenning @rustd since this is a regression (multiple times) which has also regressed OmniSharp (two times) I think we need a unit test to catch this to ensure that it doesn't happen again.

phenning:
@sayedihashimi I've sent a review out for to fix the regression. 

I think the regression happened when we did this large commit that included bower component content: ce02c500fe904e640ccc1b869223e48c74e58951. 

Large commits like this tend to be difficult to review via GitHub UI, so to mitigate, we should try to isolate bower component updates from other content updates, that will make it more likely that the content we author is reviewed more thoroughly.

Additionally, I'll look into adding a unit test to catch duplicate classes in template content.

PR #325, opened 2015-11-19
Fix regression of duplicate btn-default

phenning:
@rustd 

Fixing this for #324  

It looks like this was regressed by one of the large refactoring commits.


phenning:
Closing this PR, we'll take #323 from @scottaddie

PR #326, opened 2015-11-20
Update Web API route comment to enforce consistency

scottaddie:
The very first method in the `ValuesController.cs` file uses a colon in the comment. This is inconsistent with how the comments are formatted downstream in the same file. This PR removes the colon to enforce consistency.


rustd:
:shipit: @phenning this should be good to me merged.

PR #328, opened 2015-11-21
Fix Logging arguments

Anderman:
@rustd Change in log with arguments


Eilon:
Using C# 6 interpolated strings means that you lose the logging library's structured data. All calls to logging deliberately use what look like format strings so that the logger can log both the formatted text, but also key/value pairs so that data is searchable/sortable/filterable.

Anderman:
@Eilon You right. That  was my first proposal. But the current line won't log at all
`_logger.LogInformation(5, "User logged in with {Name} provider.", info.LoginProvider);`
It should be
`_logger.LogInformation(5, "User logged in with {0} provider.", info.LoginProvider);`

@rustd Should I make a new pull request?

Eilon:
No it's already correct - the parameters have names. It will end up logging the string `User logged in with SOME_PROVIDER provider.` and the data `Name = SOME_PROVIDER`.

Issue #327, opened 2015-11-20
rc1 default template not generating the same as beta8 [cross-post]


peterblazejewicz:
This is cross-posted feedback from [aspnet/Announcements](https://github.com/aspnet/Announcements/issues/117) project:

> (mostly) identical html/bootstrap is not rendering the same in the default template. see images. all I did was create a beta8 mvc with individual account, and the same in rc1
> ![https://cloud.githubusercontent.com/assets/15947066/11308164/dc6b3c72-8f79-11e5-90bc-35385476cb9d.png](https://cloud.githubusercontent.com/assets/15947066/11308164/dc6b3c72-8f79-11e5-90bc-35385476cb9d.png)
> ![https://cloud.githubusercontent.com/assets/15947066/11308168/dea8eade-8f79-11e5-8188-cbe23d3b2d94.png](https://cloud.githubusercontent.com/assets/15947066/11308168/dea8eade-8f79-11e5-8188-cbe23d3b2d94.png)

https://github.com/aspnet/Announcements/issues/117

/cc
@madmunsterdaddy

natelaff:
sorry I posted it there! it was a mistake.

peterblazejewicz:
The second problem with checkbox having negative offset is a known problem with tag helpers:
https://github.com/aspnet/Mvc/issues/2369
There is no clean solution for that problem yet, unless written by hand or with additional CSS (just to fix rendering) - both are not nice for templates.

rustd:
The first issue is because we switched away from bootstrap touch carousel and are using bootstrap carousel.
The second one we need to address.

peterblazejewicz:
@rustd 
- the first issue requires custom css to be added:
  https://github.com/ixisio/bootstrap-touch-carousel/blob/master/src/less/carousel.less#L88-L95
- the second issue requires custom markup, I've made one already elsewhere:
  https://github.com/peterblazejewicz/aspnet-5-bootstrap-4/blob/abeac62e0738e5a62fb42faf1b43baffeaff4608/WebApplication/Views/Account/Login.cshtml#L37-L39

I'll made two different PR, so you guys could decide.
Thanks!

PR #331, opened 2015-11-21
:bug: Fix negative offset of checkbox component

peterblazejewicz:
Related issue: #327 

This commit fixes negative offset of checkbox BS component
under BS 3.3.5 - which is now used in aspnet/Templates.
The checkbox component under 3.3.5 requires specific markup -
which at this time is not possible to render correctly with
MVC Razor tag helpers: http://git.io/v4h5M
The fix uses purpose written markup to overcome rendering
problem

The other solution to the problem is to write custom CSS rules to fix rendering problem under BS 3.3.5.
Both solutions have different drawbacks in terms of documenting platform via templates.

I've chosen custom markup as solution, as it could introduce people to some topics in MVC/Razor and because it is not hidden in CSS.

Rendered html markup:

``` html
<div class="checkbox">
    <label for="RememberMe">
        <input data-val="true" data-val-required="The Remember me? field is required." id="RememberMe" name="RememberMe" type="checkbox" value="true">
        Remember me?
    </label>
</div>
```

Thanks!

### DIFF
--- a/src/Rules/StarterWeb/IndividualAuth/Views/Account/Login.cshtml
+++ b/src/Rules/StarterWeb/IndividualAuth/Views/Account/Login.cshtml
@@ -33,8 +33,10 @@
                 <div class="form-group">
                     <div class="col-md-offset-2 col-md-10">
                         <div class="checkbox">
-                            <input asp-for="RememberMe" />
-                            <label asp-for="RememberMe"></label>
+                            <label asp-for="RememberMe">
+                                <input asp-for="RememberMe" />
+                                @Html.DisplayNameFor(m => m.RememberMe)
+                            </label>
                         </div>
                     </div>
                 </div>


rustd:
@NTaylorMullen thoughts? Can we update the TagHelpers? This is an ugly workaround on a mainline scenario?

src/Rules/StarterWeb/IndividualAuth/Views/Account/Login.cshtml
```diff
@@ -33,8 +33,11 @@
                 <div class="form-group">
                     <div class="col-md-offset-2 col-md-10">
                         <div class="checkbox">
-                            <input asp-for="RememberMe" />
-                            <label asp-for="RememberMe"></label>
+                            @* fix https://github.com/aspnet/Mvc/issues/2369 *@
```
NTaylorMullen: Nuke, we don't link to fixes in source anymore.

NTaylorMullen:
@rustd I think the change is fine. It directly reflects what's going to occur in the HTML but with some server side information added.  I wouldn't consider this a `TagHelper` issue.

NTaylorMullen:
And note: We did change in rc1 where the hidden tag gets rendered. It's now rendered at the end of the form instead of following the input tag.

src/Rules/StarterWeb/IndividualAuth/Views/Account/Login.cshtml
rustd: @NTaylorMullen Thanks! @peterblazejewicz can you please update this PR as per @NTaylorMullen comments?

peterblazejewicz: @NTaylorMullen @rustd 
Done/rebased/updated to latest dev

NTaylorMullen:
:shipit:

PR #332, opened 2015-11-22
Fix typo wrong methode FindByNameAsync=>FindByEmailAsync

Anderman:
@rustd I found that the wrong method was use for finding an email address

### DIFF
--- a/src/Rules/StarterWeb/IndividualAuth/Controllers/AccountController.cs
+++ b/src/Rules/StarterWeb/IndividualAuth/Controllers/AccountController.cs
@@ -310,7 +310,7 @@ namespace $safeprojectname$.Controllers
             {
                 return View(model);
             }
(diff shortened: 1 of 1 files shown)


rustd:
FindByName is fine since in the templates email is used as a name as well.

Anderman:
(-; Yep that's is true. I found this after changing the webapp. I think I will not be the last one.

Anderman:
@rustd Okee there are indeed more places that require that username is the same as email address.

```
                var result = await _signInManager.PasswordSignInAsync(model.Email, model.Password, model.RememberMe, lockoutOnFailure: false);
```

PasswordSignInAsync require a username. 

Next time I will first create an issue before sending a PR

rustd:
@Anderman no worries. I am closing this PR.

PR #336, opened 2015-11-23
Fix style.css file selector indentation

peterblazejewicz:
This small commit fixes a selector declaration indentation,
preventing from propagating to next lines - which is
most IDE/editors default behavior
Thanks!

### DIFF
--- a/src/BaseTemplates/StarterWeb/wwwroot/css/site.css
+++ b/src/BaseTemplates/StarterWeb/wwwroot/css/site.css
@@ -18,7 +18,7 @@ textarea {
 }
 
 /* Carousel */
-    .carousel-caption p {
-        font-size: 20px;
-        line-height: 1.4;
-    }
+.carousel-caption p {
+    font-size: 20px;
+    line-height: 1.4;
+}


rustd:
@peterblazejewicz I don't think these classes are needed since we are not using the bootstrap touch carousel library. Can you please see if the images look fine without this css and remove it if not needed.

peterblazejewicz:
@rustd 

> I don't think these classes are needed since we are not using the bootstrap touch carousel library. Can you please see if the images look fine without this css and remove it if not needed.

They are only for to visually bump paragraph text size and they are similar to other similar design decision:
https://github.com/twbs/bootstrap/blob/master/docs/examples/carousel/carousel.css#L118-L123

I think the project carousel requires simply:
- corrected markup to align text and button
- a hidden content on small screens:

> 768px
> 

< 768px


The changes are minimal and only in html markup using BS responsive class utils. 
Let me know if this is OK for other PR.

This PR is just code formatting thing.
Thanks!

rustd:
Yes, we still need the CSS changes so let's only take the code formatting changes.

rustd:
@phenning let's take only the code formatting changes to the css file.

PR #337, opened 2015-11-24
:art: Two small fixes to landing page carousel

peterblazejewicz:
This commit introduces two different fixes to home page
carousel component layout:
- the interactive elements are wrapped into paragraph tag and
  are now displayed in separate row in a block
- the carousel caption paragraphs are now hidden on small
  screens
  This changes are quick and do not require changes in CSS style.
  They should be changed in the future when content design will be
  under review

The > 768 and < 768 screens after this changes:

Note: that is somehow obvious that design was not mobile-first when first created. The small screens landing page needs some love (the content should be higher and with different background images for carousel (example comes with image set via `object-fit: fill`, so the content is distorted):

image

For related subject and  discussion, please see:
https://github.com/aspnet/Templates/pull/336#issuecomment-159356581

Thanks!


rustd:
The code was there to show the text. Let's not take this PR.

PR #339, opened 2015-11-25
Fix appsetting documentation in StarterWeb template

nathanchere:
Previously redirecting to MS home page.
Now points to correct redirect.


rustd:
:shipit: @phenning this is good to be merged.

PR #349, opened 2015-12-09
Update Project Layout

### LINKED ISSUE #290: Update Project Layout
Update Project Layout
Delete the ViewModels folder and move the ViewModel files under the models folder. Also update the namespaces in the ViewModel classes to reflect this.
Create a data folder and move the migrations folder under it.  This depends on https://github.com/aspnet/EntityFramework/issues/2709 but we can update the templates for the ViewModels change first

Models
--AccountViewModels
----C# files
--ManageViewModels
----C# files

Data
--Migrations folder here

phenning:
Closes #290 
@rustd , @Eilon @davidfowl


Eilon:
Assuming everything still works, :shipit: 

@DamianEdwards FYI also

rustd:
As part of this bug please create a Data folder and move the Migrations over there.

phenning:
The bug was updated to not do that. Do we want to that now?

"We don't need EF changes. EF will follow wherever the migrations are"

phenning:
In regards to my last comment, I think you meant we don't need EF runtime changes. I misread the bug.  I'll move the migrations into a "Data" folder.

phenning:
Merged to dev

rustd:
Thanks! We don't need any EF changes for this.

PR #356, opened 2015-12-15
Web API: Windows Auth Template

### LINKED ISSUE #202: Web API: Windows Auth Template

phenning:
Closes #202 

@rustd, @joeloff

### DIFF
--- a/Templates.Settings.targets
+++ b/Templates.Settings.targets
@@ -83,7 +83,9 @@
         <TestTemplates Include="$(TemplatesBin)\test\ConsoleApp"/>
         <TestTemplates Include="$(TemplatesBin)\test\EmptyWeb"/>
         <TestTemplates Include="$(TemplatesBin)\test\WebAPI"/>
+        <TestTemplates Include="$(TemplatesBin)\test\WebAPI.WindowsAuth"/>
         <TestTemplates Include="$(TemplatesBin)\test\WebAPI.AI"/>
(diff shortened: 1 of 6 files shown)


brockallen:
I don't really see the contents of the template, so is this a browser-based template? If so I wanted to ask if the template include XSRF protections (since it says you're using Windows Auth).

phenning:
This template has no UI, we just need to update the launchSettings to enable windows auth and disable anonymous.

This follows the same pattern as StarterWeb Windows auth, only that template is a UI template so it also updates the _Layout.cshtml.

brockallen:
Ok... so then perhaps a comment about XSRF protection on Web APIs might be useful (especially when using an implicit authentication mechanism like windows authN).

phenning:
@rustd 

Should we provide a ValuesController for all auth types with:
1. [Authorize] set at the class level.
2. Add a comment in Post, Put and Delete:

// For more information on protecting this API from Cross Site Request Forgery (CSRF) attacks, see [fwlink].

rustd:
@phenning let's do it. My sample had a ValuesController as well.
/cc @danroth27

brockallen:
I suspect the link you will be providing should be docs from none other than @blowdart. :trollface:

phenning:
Updated the PR.

rustd:
:shipit:

PR #358, opened 2015-12-17
Fix invalid jquery-validation _references.js path

scottaddie:
The path defined for the jquery-validation library is invalid in the `_references.js` file. As a result, the following message is displayed in the Output window's "JavaScript Language Service" view: "Referenced file 'lib/jquery-validation/jquery.validate.js' not found."


phenning:
Thanks!

Issue #168, opened 2015-08-27
Usability: Hide "bower.json", ".bowerrc" and "package.json" from project


madskristensen:
The feature in the project system that enables this is being designed right now. I'll update this issue with how to hide the files when the feature has been designed.

phenning:
28b74daba116d50c15debc10c7c8d55dead583a7

PR #371, opened 2016-01-07
Add .npmrc to DnxInvisibleContent item group

scottaddie:
The `DnxInvisibleContent` item group which was added to the project file (per issue https://github.com/aspnet/Templates/issues/168) didn't take into consideration the scenario when the developer adds a `.npmrc` file to the project. This PR adds this file to the item group. The thought is that the `.npmrc` file will receive the same treatment as the `.bowerrc` file. In my opinion, it looks odd when the `.npmrc` file is visible in Solution Explorer but the `package.json` file is invisible. These 2 npm artifacts should receive the same visibility treatment, as is the case with the 2 Bower artifacts.

### DIFF
--- a/src/BaseTemplates/EmptyWeb/EmptyWeb.xproj
+++ b/src/BaseTemplates/EmptyWeb/EmptyWeb.xproj
@@ -20,6 +20,7 @@
     <DnxInvisibleContent Include="bower.json" />
     <DnxInvisibleContent Include=".bowerrc" />
     <DnxInvisibleContent Include="package.json" />
+    <DnxInvisibleContent Include=".npmrc" />
   </ItemGroup>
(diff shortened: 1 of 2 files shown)


rustd:
@madskristensen thoughts?

madskristensen:
Do we shop a .npmrc file in the templates? If not, I don't think we should change this

scottaddie:
@madskristensen To my knowledge, the `.npmrc` file isn't included in the templates. My thought was that the `DnxInvisibleContent` item group seems to be treated as a blacklisting of files to hide in Solution Explorer. In other words, if you delete the `package.json` file from the project, the corresponding `DnxInvisibleContent` node remains in the project file.

madskristensen:
In that case it makes sense to me. Thanks for clarifying

scottaddie:
@madskristensen It sounds like we're all in agreement then. Is this safe to merge?

madskristensen:
:+1:

PR #372, opened 2016-01-07
Add ReturnUrl persistence to Individual Auth template

### LINKED ISSUE #359: Return URL persistence in ASP.NET 5 web application template with individual accounts
The template does not currently persist any return URL information as you navigate around registering for a new account, for example. Return URLs are standard in the login process. The links and forms need to be updated to persist this data, one such solution is outlined in my blog post here:

http://overengineer.net/Using-OpenIddict-to-easily-add-token-authentication-to-your-.NET-web-apps/#almost-done

phenning:
Closes #359 

@rustd


rustd:
:shipit:

PR #374, opened 2016-01-08
Add aspnet_useplatformhandler token for certain template scenarios.

phenning:
@BillHiebert @rustd 

Adding $if$ logic, need to strip this out for the dnu build to verify the closure.


phenning:
merged

Issue #373, opened 2016-01-08
Minor formatting issues with project.json and _references.js in some templates


balachir:
Minor formatting issues with project.json and _references show up when checking out project from source control.

project.json
![image](https://cloud.githubusercontent.com/assets/8246794/12209827/85c354b0-b60a-11e5-97ef-0111ac29622e.png)

_references.js
![image](https://cloud.githubusercontent.com/assets/8246794/12209823/7da4e604-b60a-11e5-8385-0b23fcaee0cf.png)

rustd:
Closing as fixed

PR #375, opened 2016-01-08
Fix minor formatting issues with project.json and _references.js 

balachir:
Fix for https://github.com/aspnet/Templates/issues/373


phenning:
:shipit:

PR #403, opened 2016-01-27
Branding changes

phenning:
@rustd 

Updated the usersecretsId as well to remove aspnet5 from the id pattern

### DIFF
--- a/src/BaseTemplates/DNXEmptyWeb.vstemplate
+++ b/src/BaseTemplates/DNXEmptyWeb.vstemplate
@@ -1,8 +1,8 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <VSTemplate Version="3.0.0" Type="ProjectGroup" xmlns="http://schemas.microsoft.com/developer/vstemplate/2005" xmlns:sdk="http://schemas.microsoft.com/developer/vstemplate-sdkextension/2010">
   <TemplateData>
-    <Name>ASP.NET 5 Empty Web Application</Name>
-    <Description>A project for creating an empty ASP.NET 5 web application</Description>
+    <Name>ASP.NET Core 1 Empty Web Application</Name>
+    <Description>A project for creating an empty ASP.NET Core 1 web application</Description>
     <Icon>Empty.png</Icon>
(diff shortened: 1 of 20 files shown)


rustd:
:shipit: 
As a followup should we change the vs template file names from DNX to DotNet or something else to match the grouping parameter @joeloff is using for Item templates. We would only do this if this is something a customer would see.

joeloff:
Customers don't typically see these file names as they live in the VS install folders so it shouldn't be a customer facing issue I think

phenning:
I think this is something we could change for RTW though, if we wanted to.
- Phil

On Jan 27, 2016, at 6:52 PM, joeloff <notifications@github.com<mailto:notifications@github.com>> wrote:

Customers don't typically see these file names as they live in the VS install folders so it shouldn't be a customer facing issue I think

##

rustd:
I see. This is fine then.
/cc @Eilon JFYI

src/BaseTemplates/DNXEmptyWeb.vstemplate
```diff
@@ -1,8 +1,8 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <VSTemplate Version="3.0.0" Type="ProjectGroup" xmlns="http://schemas.microsoft.com/developer/vstemplate/2005" xmlns:sdk="http://schemas.microsoft.com/developer/vstemplate-sdkextension/2010">
   <TemplateData>
-    <Name>ASP.NET 5 Empty Web Application</Name>
-    <Description>A project for creating an empty ASP.NET 5 web application</Description>
+    <Name>ASP.NET Core 1 Empty Web Application</Name>
```
Eilon, line 4: If these names show up in the New Project dialog then these shouldn't have versions in them.

cc @DamianEdwards

rustd, line 4: They will show up as description for the template as shown below.

oneaspnet

Eilon, line 4: Yeah I think @DamianEdwards doesn't want versions there... @DamianEdwards can you confirm?

Issue #369, opened 2016-01-05
Invalid package.json name property value


scottaddie:
Per the `package.json` file's [property specs](https://github.com/npm/validate-npm-package-name#naming-rules), the `name` property's value cannot contain uppercase letters. The quickest way to test this is to open a command shell and run `npm init`:

![package_json_name](https://cloud.githubusercontent.com/assets/10702007/12124220/73fd76f0-b3aa-11e5-9396-7a359d89aa26.png)

The default value in the ASP.NET 5 templates is "ASP.NET", which violates the spec. I can submit a PR to address the project templates, but are there any preferences on a new value?

The "NPM Configuration File" item template will also need to be changed to reflect this new value, as it currently contains this:

``` json
{
    "version": "1.0.0",
    "name": "ASP.NET",
    "private": true,
    "devDependencies": {
    }
}
```

scottaddie:
@madskristensen Yet another piece to this proposed change is to enforce lowercase via a pattern property regex in the SchemaStore repo. Thoughts?

madskristensen:
Makes sense

PR #408, opened 2016-02-05
Alter package.json name value to meet specification

scottaddie:
This PR addresses issue https://github.com/aspnet/Templates/issues/369. By changing the `name` property value to all lowercase, npm's specification is met.

Note that the "NPM Configuration File" item template suffers from the same problems and should be updated too.

### DIFF
--- a/src/BaseTemplates/StarterWeb/package.json
+++ b/src/BaseTemplates/StarterWeb/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "ASP.NET",
+  "name": "asp.net",
   "version": "0.0.0",
   "private": true,
   "devDependencies": {


scottaddie:
@madskristensen Can you please review this PR when you get a chance? The associated SchemaStore repo change was merged already, so it would be nice to get this in to eliminate schema validation warnings.

phenning:
@rustd, I'm assuming functionality trumps branding in this case.

madskristensen:
Wouldn't the restriction on lowercase characters only be there if `private` hasn't been set to `true`? If so, couldn't the schema then enforce that rule instead of changing it here?

scottaddie:
This does become a much bigger deal when `private` is set to `false`; however, npm still enforces it everywhere. Open a command shell, and run `npm init`. This command creates the `package.json` file. Notice that the very first question asks for a name. Using uppercase letters at this point leads to an error message of "Sorry, name can no longer contain capital letters." and doesn't allow you to proceed.

FYI, here are the package name validation rules:
https://github.com/npm/validate-npm-package-name#naming-rules

It used to be acceptable to use uppercase letters in really old versions of npm. My understanding is that the npm dependency which is installed with VS 2015 Update 1 will be updated in Update 2. It would then seem to make sense that the specification is adopted to support this newer version's demands.

madskristensen:
Thanks for the clarification @scottaddie 

This change seems like one that would be foolish not to take. :shipit:

scottaddie:
@madskristensen FYI, this will also need to be changed in the Apache Cordova project template and the "NPM Configuration File" item template. I don't believe that either of these are open source, so this will need to be routed to the appropriate teams.

madskristensen:
@phenning Can you make the change to the item template as well, please?

phenning:
@madskristensen Will do. Merging this change now.

madskristensen:
Awesome. Thanks @phenning and @scottaddie

PR #409, opened 2016-02-09
Modifying project.json templates to run publish-iis after publish

### LINKED ISSUE #384: Update templates for dotnet tools publish changes

moozzyk:
Fixes #384

### DIFF
--- a/src/BaseTemplates/EmptyWeb/Program.cs
+++ b/src/BaseTemplates/EmptyWeb/Program.cs
@@ -15,6 +15,7 @@ namespace $safeprojectname$
             var hostingConfiguration = WebApplicationConfiguration.GetDefault(args);
 
             var application = new WebApplicationBuilder()
+                .UseServerFactory("Microsoft.AspNet.Server.Kestrel")
                 .UseConfiguration(hostingConfiguration)
                 .UseStartup<Startup>()
                 .Build();
--- a/src/BaseTemplates/WebAPI/project.json
+++ b/src/BaseTemplates/WebAPI/project.json
@@ -25,12 +25,26 @@
     "dnxcore50": { }
   },
 
-  "exclude": [
+  "tools": {
+    "dotnet-publish-iis": "1.0.0-*"
+  },
+
+  "content": [
     "wwwroot",
+    "Views",
+    "appsettings.json"
+  ],
+
+  "exclude": [
     "node_modules"
   ],
+
   "publishExclude": [
     "**.user",
     "**.vspscc"
-  ]
+  ],
+
+  "scripts": {
+    "postpublish": [ "dotnet publish-iis --publish-folder %publish:OutputPath%" ]
+  }
 }
--- a/src/Rules/StarterWeb/OrganizationalAuth/Common/project.json
+++ b/src/Rules/StarterWeb/OrganizationalAuth/Common/project.json
@@ -33,15 +33,27 @@
     "dnxcore50": { }
   },
 
-  "exclude": [
+  "tools": {
+    "dotnet-publish-iis": "1.0.0-*"
+  },
+
+  "content": [
     "wwwroot",
+    "Views",
+    "appsettings.json"
+  ],
+
+  "exclude": [
     "node_modules"
   ],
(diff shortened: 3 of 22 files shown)


src/BaseTemplates/EmptyWeb/Program.cs
```diff
@@ -15,6 +15,7 @@ public static void Main(string[] args)
             var hostingConfiguration = WebApplicationConfiguration.GetDefault(args);
 
             var application = new WebApplicationBuilder()
+                .UseServerFactory("Microsoft.AspNet.Server.Kestrel")
```
moozzyk, line 18: The recommended way is to have this in code instead of using the hosting.json file. I can revert this if you prefer to keep hosting.json

rustd, line 18: We don't need hosting.json file

rustd:
@phenning @BillHiebert FYI

src/BaseTemplates/EmptyWeb/Program.cs
moozzyk, line 18: :+1:

phenning:
I assume we'll also need a coherence-signed from this year, or should this work with build #15697?

rustd:
@mlorbetske FYI

moozzyk:
@phenning - not sure how I missed the the Rules subfolder. Will fix these as well. 
With regards to whether you need a new build I think it depends on what you are planning to do. I intentionally have not moved packages/namespaces to AspNetCore so that you can still build and run tests. The "tools" section is new and dnx/dnu does not understand it (and therefore skips it). You need to use `dotnet` to see it work. So, while the templates can be build and package the projects that they are generated may work differently depending on whether you use `dnx` or `dotnet`

moozzyk:
:up::date:

moozzyk:
@phenning @rustd - I noticed that the templates in the Rules don't have hosting.json files and are not setting the webserver programmatically. I don't remember if it was not required before but I think going forward setting a webserver is required...

phenning:
the files in the rules folder overwrite existing files in the base template, so they would have inherited the base templates' hosting.json and Program.cs, so the only changes needed are the changes to the project.json files.

src/Rules/WebAPI/OrganizationalAuth/Single/project.json
```diff
@@ -28,12 +28,20 @@
     "dnxcore50": { }
   },
 
+  "content": [
+    "wwwroot"
+  ],
```
phenning, line 35: appsettings.json and Views?

src/Rules/StarterWeb/OrganizationalAuth/Common/project.json
```diff
@@ -33,15 +33,26 @@
...
   },
 
-  "exclude": [
+  "tools": {
+    "dotnet-publish-iis": "1.0.0-*"
+  },
+
+  "content": [
     "wwwroot",
+    "Views"
```
phenning: appsettings.json needed here?

src/BaseTemplates/WebAPI/project.json
```diff
@@ -25,12 +25,25 @@
...
   },
 
-  "exclude": [
+  "tools": {
+    "dotnet-publish-iis": "1.0.0-*"
+  },
+
+  "content": [
     "wwwroot",
+    "appsettings.json"
```
phenning, line 35: No Views? Its in the rules for WebAPI, below

moozzyk, line 35: OK. Honestly I don't know how you put together the complete templates and what the relation between BaseTemplates and Rules is. If there was no Views or appSetting.json in the folder where the project.json file lives I left it out. 

I will try to figure this out and update the PR again.

moozzyk:
@phenning Thanks for explaining how this works - now I can see how the rules are defined in the Template.xml. I will update the PR accordingly.

moozzyk:
:up::date: # 2

phenning:
:+1: :shipit:

PR #413, opened 2016-02-19
Use resources from the dotnet package

rchande:
The Class Library (.NET Core) and Console App (.NET Core) vstemplates
should load their resources from the dotnet VS package so that the correct
text is displayed even if WTE aren't installed.


rchande:
I noticed while testing my earlier string change that the resources aren't loaded if you don't install the web tools. After this change I'll update the templates on the VS side and move the resources to the dotnet VS package.

Please review @phenning @cartermp @rustd @DamianEdwards

cartermp:
:shipit:

rchande:
Ping? I'd like to get this merged in so I can do the VS side.

phenning:
Merged

rchande:
Thanks @phenning! When will there be a new templates build for me to insert into VS?

PR #414, opened 2016-02-23
Point the correct vstemplates at the dotnet package

rchande:
I changed "classlibrary.vstemplate" and "consoleapp.vstemplate" to reference the dotnet package when I should have changed "dnxclasslibrary.vstemplate" and "dnxconsoleapp.vstemplate". The dnx*.vstemplates are actually ones that are dropped under .NET Core.


rchande:
I tested this by building setup with these vstemplates and my VS-side changes and now the correct strings are picked up.

rchande:
Please review @cartermp @phenning . Thanks!

PR #415, opened 2016-02-23
React to Connection string config changes

HaoK:
cc @divega @danroth27 @phenning 

Let me know if you want me to just merge this

### DIFF
--- a/src/Rules/StarterWeb/AI/IndividualAuth/appsettings.json
+++ b/src/Rules/StarterWeb/AI/IndividualAuth/appsettings.json
@@ -2,7 +2,9 @@
   "ApplicationInsights": {
     "InstrumentationKey": ""
   },
-  "DefaultConnection": "Server=(localdb)\\mssqllocaldb;Database=aspnet-$safeprojectname$-$guid2$;Trusted_Connection=True;MultipleActiveResultSets=true",
+  "ConnectionStrings": {
+    "DefaultConnection": "Server=(localdb)\\mssqllocaldb;Database=aspnet-$safeprojectname$-$guid2$;Trusted_Connection=True;MultipleActiveResultSets=true"
+  },
   "Logging": {
     "IncludeScopes": false,
     "LogLevel": {
--- a/src/Rules/StarterWeb/AI/IndividualAuth/appsettings.json
+++ b/src/Rules/StarterWeb/AI/IndividualAuth/appsettings.json
@@ -2,11 +2,7 @@
   "ApplicationInsights": {
     "InstrumentationKey": ""
   },
-  "Data": {
-    "DefaultConnection": {
-      "ConnectionString": "Server=(localdb)\\mssqllocaldb;Database=aspnet-$safeprojectname$-$guid2$;Trusted_Connection=True;MultipleActiveResultSets=true"
-    }
-  },
+  "DefaultConnection": "Server=(localdb)\\mssqllocaldb;Database=aspnet-$safeprojectname$-$guid2$;Trusted_Connection=True;MultipleActiveResultSets=true",
   "Logging": {
     "IncludeScopes": false,
     "LogLevel": {
--- a/src/Rules/StarterWeb/AI/IndividualAuth/Startup.cs
+++ b/src/Rules/StarterWeb/AI/IndividualAuth/Startup.cs
@@ -52,7 +52,7 @@ namespace $safeprojectname$
             services.AddEntityFramework()
                 .AddSqlServer()
                 .AddDbContext<ApplicationDbContext>(options =>
-                    options.UseSqlServer(Configuration["DefaultConnection"]));
+                    options.UseSqlServer(Configuration.GetConnectionString("DefaultConnection")));
 
             services.AddIdentity<ApplicationUser, IdentityRole>()
                 .AddEntityFrameworkStores<ApplicationDbContext>()
--- a/src/Rules/StarterWeb/AI/IndividualAuth/Startup.cs
+++ b/src/Rules/StarterWeb/AI/IndividualAuth/Startup.cs
@@ -52,7 +52,7 @@ namespace $safeprojectname$
             services.AddEntityFramework()
                 .AddSqlServer()
                 .AddDbContext<ApplicationDbContext>(options =>
-                    options.UseSqlServer(Configuration["Data:DefaultConnection:ConnectionString"]));
+                    options.UseSqlServer(Configuration["DefaultConnection"]));
 
             services.AddIdentity<ApplicationUser, IdentityRole>()
                 .AddEntityFrameworkStores<ApplicationDbContext>()
--- a/src/Rules/StarterWeb/IndividualAuth/Startup.cs
+++ b/src/Rules/StarterWeb/IndividualAuth/Startup.cs
@@ -46,7 +46,7 @@ namespace $safeprojectname$
             services.AddEntityFramework()
                 .AddSqlServer()
                 .AddDbContext<ApplicationDbContext>(options =>
-                    options.UseSqlServer(Configuration["DefaultConnection"]));
+                    options.UseSqlServer(Configuration.GetConnectionString("DefaultConnection")));
 
             services.AddIdentity<ApplicationUser, IdentityRole>()
                 .AddEntityFrameworkStores<ApplicationDbContext>()
--- a/src/Rules/StarterWeb/IndividualAuth/Startup.cs
+++ b/src/Rules/StarterWeb/IndividualAuth/Startup.cs
@@ -46,7 +46,7 @@ namespace $safeprojectname$
             services.AddEntityFramework()
                 .AddSqlServer()
                 .AddDbContext<ApplicationDbContext>(options =>
-                    options.UseSqlServer(Configuration["Data:DefaultConnection:ConnectionString"]));
+                    options.UseSqlServer(Configuration["DefaultConnection"]));
 
             services.AddIdentity<ApplicationUser, IdentityRole>()
                 .AddEntityFrameworkStores<ApplicationDbContext>()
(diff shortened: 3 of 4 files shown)


rustd:
What's the motivation for this?

HaoK:
https://github.com/aspnet/Configuration/issues/376

rustd:
@sayedihashimi FYI in case this affects publishing to Azure App Service
@HaoK you would also need to update Startup where the connectionstring is accessed by EF

HaoK:
Good catch on startup, I only grep'd json, my bad... Updated

src/Rules/StarterWeb/AI/IndividualAuth/appsettings.json
```diff
@@ -2,11 +2,7 @@
   "ApplicationInsights": {
     "InstrumentationKey": ""
   },
-  "Data": {
-    "DefaultConnection": {
-      "ConnectionString": "Server=(localdb)\\mssqllocaldb;Database=aspnet-$safeprojectname$-$guid2$;Trusted_Connection=True;MultipleActiveResultSets=true"
-    }
-  },
+  "DefaultConnection": "Server=(localdb)\\mssqllocaldb;Database=aspnet-$safeprojectname$-$guid2$;Trusted_Connection=True;MultipleActiveResultSets=true",
```
rustd: Can we keep the data section? It was something we had agreed upon earlier when reviewing templates. How  about something like the following?

``` C#
"Data" :{ "WebAppNameConnectionString":"<value>"}
or
"Data" :{ "ConnectionString":"<value>"}
```

divega: I understand that we had agreed previously on having the "Data" section but I think it means much less right now. 

The change that just went in Configuration is about removing the "Data:MyConnection:ConnectionString" convention we used to have and map to Azure Web Site connection strings. Now connection strings in Azure Web Sites will map directly to settings with the same name. See https://github.com/aspnet/Configuration/issues/376 for more context. This means we are no longer mangling the user provided name and gives users more freedom to structure their application settings in whatever way they want.

This also means that for customers to match "Data:ConnectionString" they will have to type exactly "Data:ConnectionString" in the Azure Web Site UI.

Personally I don't see much value in keeping the "Data" section in the default configuration in the template given that it sets precedent of a convention that is actually superfluous and unnecessary, but in theory we could keep it.

cc @Eilon @danroth27 What do you guys think? Anyone else needs to review this?

divega: cc @DamianEdwards in case he wants to weigh in.

sayedihashimi:
Adding @bradygaster

src/Rules/StarterWeb/AI/IndividualAuth/appsettings.json
shirhatti: @rustd Weighing in.
The name you enter in the portal should be the name you use to access the connection string. 

After aspnet/Configuration#376, the only way to accommodate the `Data` section in the templates is have customers type `Data:MyConnection:ConnectionString` into the portal, which is just **awful**.

Unless I am missing context from why previously agreed to the `data` section, we should merge these changes in.

danroth27: Why not simply call the key ConnectionString instead of DefaultConnection? What does "default" even mean in this context?

bradygaster-zz: "DefaultConnection" is the connection string name developers have come to expect from the FNP experience. We also added UI to the SQL server/db-creation process in SDK 2.8.2 to re-surface the connection string per customer requests to make the experience more like the pre-2.8.1 experience. Users _expect_ to see the name of their DB connection string as DefaultConnection

bradygaster-zz: Also, agreed with @shirhatti on the verbosity of the connection string.It isn't pretty to have to type all that - making any assumptions about the nomenclature users "should use" for their connection string is kind of a bad proposition. I'm also chatting with the App Service folks to make sure there aren't any assumptions made on the names of the connection strings. We should make as few assumptions as possible about prefixes and so on.

danroth27: Makes sense - DefaultConnection it is then.

phenning:
Looks good to me, I agree with the more user friendly naming of DefaultConnection for Azure customers.

phenning:
Are we OK taking this now?

bradygaster-zz:
Adding @mlorbetske to the conversation. I'm okay with it but we're in verifying there won't be any negative impacts with the App Service team. Will want to discuss with Mike before we close.

divega:
I would really like to help get this to closure, but I am having second thoughts on the whole change :smile:

Looking at the current default appsettings.json file I realized that we have a few sections already like `Logging` and `AppInsights`. Rather than having `DefaultConnection` appear as a top level key I am starting to like the idea of renaming `Data` to `ConnectionStrings` and (still) remove the unnecessary extra `:ConnectionString` level. E.g. in this file, it would look like this:

``` JSON
 "ConnectionStrings": {  
    "DefaultConnection": "Server=(localdb)\\mssqllocaldb;Database=aspnet-$safeprojectname$-$guid2$;Trusted_Connection=True;MultipleActiveResultSets=true"  
  },  
```

Besides having all the connection strings together in the JSON file another aspect I like about this is that we could eventually add bespoke extension methods for `IConfigurationRoot` to retrieve these well known sections in code, e.g.

``` C#
var defaultConnection = config.GetConnectionStrings()["DefaultConnection"];
```

I think it was @davidfowl who suggested we added something like that some time ago, and I guess I am finally coming to terms with it :smile:

HaoK:
Yay for `config.GetConnectionString("DefaultConnection")`

rustd:
I like the change to appsettings to have a section for connectionStrings. It could be used to add more connections to Storage and other data sources.
I am neutral on adding an API to get the connectionstring value.
+1 on taking in this change

bradygaster-zz:
+1

divega:
> Yay for  config.GetConnectionString("DefaultConnection") 

I think you have misread what I wrote :smile: (I actually missed the parenthesis, my bad).

divega:
`GetConnectionStrings()` would be just sugar for `GetSection("ConnectionStrings")`.

divega:
If people generally agree (I already talked to @rowanmiller and he is ok with it, I would like to hear from @danroth27 and others). Then we need to make the runtime change first.

danroth27:
@divega @bradygaster Doesn't this effect publishing to Azure? I thought Azure publishing specifically creates a "DefaultConnection" environment variable.

divega:
@danroth27 the new proposal doesn't change how the environment variable's name (it is some provider prefix + connection string name). But maybe I am missing something.

bradygaster-zz:
@danroth27 @divega @mlorbetske tooling does default the connection string name being dropped in as "DefaultConnection." If tooling needs to change to support this we can change, but at the end of the day, we need to set the connection string name the customer _wants_ and not default it. our default was just to unbreak users who had issues with the deprecation of the automatic setting in SDK 2.8.1. We added DefaultConnection back in 2.8.2

danroth27:
Let's not break the publish tooling unless we have a solid plan on how the team can continue to test publish to azure scenarios and how it will get fixed for customers.

bradygaster-zz:
@danroth27 to clarify (just tested with the current bits). Whatever the customer provides for the name of the Connection String on the SQL provisioning dialog during App Service creation, is whatever gets dropped in as the name of the connection string in Azure. So if a user has this in their app settings JSON file (given the RC1 tooling and Azure SDK):

```
"Data": {
    "PersonConnection": {
      "ConnectionString": "Server=(localdb)\\mssqllocaldb;Database=PeopleDb;Trusted_Connection=True;MultipleActiveResultSets=true"
    }
  },
```

The name of the connection string they'll get when they publish their app to Azure is "PersonConnection." They'd then use code like this in their C# to get the connection string in the Startup.cs file. 

```
var connection = Configuration["Data:PersonConnection:ConnectionString"];
            services.AddEntityFramework()
                .AddSqlServer()
                .AddDbContext<PersonContext>(options => options.UseSqlServer(connection));
```

Looks like the templates need to be adjusted to support the shorter connection string name, but for Azure, we set whatever the user specifies.

divega:
@bradygaster Are you saying that a configuration key called "Data:PersonConnection:ConnectionString" gets translated into a connection string called "PersonConnection"? Then it seems that the previous rules we had here in the runtime in RC1 are already hardcoded into the tooling and our hands are tied. Let's meet offline to sort this out.

bradygaster-zz:
That's correct, @divega. Also verified with @davidebbo that the prefixing App Service is doing shouldn't impact this. According to @davidebbo "a connection string is a connection string" and the prefix nomenclature they're implementing shouldn't impact this story.

HaoK:
Updated with proposal from https://github.com/aspnet/Configuration/issues/406

Looks a bit wordy to me, perhaps DefaultConnection => Default would be better given the nesting

rustd:
@HaoK @divega What is the current state of this PR? Has this PR been updated for the final design?

HaoK:
The current proposed design is what's in the PR I believe. To switch to using config.GetConnectionString instead of directly accessing config. If everyone is happy I can make the runtime changes.

rustd:
I am fine with the change as long as this works for publishing to App Service. Default is better than DefaultConnection @divega you good with this?

HaoK:
The runtime changes were checked in to Configuration with 9230cdb59ad38b74435bc39c0c8699d6a07c5e39

rustd:
Thanks! Can you please update this PR to match the runtime changes.

HaoK:
There's nothing to update

rustd:
@phenning This can be merged with the next successful coherence signed build. Please make sure that when you publish the Web App to Azure you can register/ login

PR #416, opened 2016-02-26
Update OutputPath in xproj files.

phenning:
Dotnet already adds the project directory, configuration, rid and tfm to whatever base path we set.. We can leave the intermediate output directory alone.

@BillHiebert


phenning:
Fixes #417

PR #419, opened 2016-03-01
Update .NET Core templates

rchande:
Use netstandard TFMs
Update to most recent package build, which allows exclusion of imports statements.

### DIFF
--- a/src/BaseTemplates/ClassLibrary/project.json
+++ b/src/BaseTemplates/ClassLibrary/project.json
@@ -2,10 +2,10 @@
     "version": "1.0.0-*",
 
     "dependencies": {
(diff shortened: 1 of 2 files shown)


rchande:
Please review @ericstj @cartermp @phenning

Pilchie:
:+1: from me.

cartermp:
LGTM

ericstj:
23830 packages aren't published yet, won't that be a problem?  Also, have you coordinated with @jhendrixMSFT to make sure you use the same package versions are inserted into VS.

ericstj:
And actually the package versions would be 23901 since the month changed.

PR #424, opened 2016-03-04
Add UseForwardedHeaders to templates

phenning:
Also removes conditional inclusion of the IISPlatformHandler package, since this is now needed for all templates due to the code in program.cs

@Tratcher , @rustd , @davidfowl

### DIFF
--- a/src/BaseTemplates/EmptyWeb/Startup.cs
+++ b/src/BaseTemplates/EmptyWeb/Startup.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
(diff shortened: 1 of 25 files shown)


src/BaseTemplates/EmptyWeb/Startup.cs
```diff
@@ -22,8 +23,13 @@ public void Configure(IApplicationBuilder app)
         {$if$ ($aspnet_useplatformhandler$ == false)
```
Tratcher, line 23: Do you still need this conditional?

phenning, line 23: I was leaning toward leaving the code conditionals in there for the service fabric scenario, so that the code doesn't even get invoked.

Tratcher:
:+1: though @DamianEdwards had some comments last this came up.

phenning:
Merging this now.

PR #426, opened 2016-03-08
Use dotnet-ef

### LINKED ISSUE #364: EF command in RC2 in Individual Auth template.
See comment here:
https://github.com/aspnet/Templates/pull/363/files#r48389518

"The reason why we removed multiple commands in our samples is to prepare for the move towards dotnet cli where commands don't exist. Maybe we should do the same here and remove ef ."

bricelam:
Resolves #364


rustd:
@phenning this looks good to me and it will address #364

rustd:
@mlorbetske FYI

phenning:
:shipit:

PR #430, opened 2016-03-09
Remove role="form" from cshtml pages.

### LINKED ISSUE #429: W3C markup validation error
As per https://validator.w3.org/ we should remove role=form from all form tags. This shows up as warning

phenning:
Fixes #429 

@rustd


rustd:
:shipit:

PR #431, opened 2016-03-09
Add Development environment info to error page

### LINKED ISSUE #351: Add information about how to enable development mode to Error.cshtml
Development mode (`IHostingEnvironment.IsDevelopment()`) is only set by default when running from VS. This means if you run from OSX/Linux (without setting the environment variable) you just get Error.cshtml rather than our more helpful error pages.

We should add some info to Error.cshtml instructing folks how to enable development mode if they want to see more info (with a warning not to do it in production). This is a similar idea to what we used to do with the yellow screen of death.

rowanmiller:
If you are not running from VS, then by default you are not in
development environment (until you set an environment variable). Adding
some info to the non-Development error page to help out folks developing
outside of VS.

Resolves #351 

cc @coolcsh for his input on whether the text would have helped him, since he ran into this prepping for a demo

### DIFF
--- a/src/BaseTemplates/StarterWeb/Views/Shared/Error.cshtml
+++ b/src/BaseTemplates/StarterWeb/Views/Shared/Error.cshtml
@@ -4,3 +4,11 @@
 
 <h1 class="text-danger">Error.</h1>
 <h2 class="text-danger">An error occurred while processing your request.</h2>
+
+<h3>Development Mode</h3>
+<p>
+    Swapping to <strong>Development</strong> environment will display more detailed information about the error that occurred.
+</p>
+<p>
+    <strong>Development environment should not be enabled in deployed applications</strong>, as it can result in sensitive information from exceptions being displayed to end users. For local debugging, development environment can be enabled by setting the <strong>ASPNETCORE_ENVIRONMENT</strong> environment variable to <strong>Development</strong>, and restarting the application.
+</p>


phenning:
Any feed back on this? @rustd, @sayedihashimi

rustd:
I had a chat with @rowanmiller The simplest change is to change the Error page to show a message always saying that switching to development environment will give you more details.

Here is what the page would look like?

``` html
    <p>
        Swapping to <strong>Development</strong> environment will display more detailed information on the error that occurred.
    </p>
    <p>
        Development environment can be enabled by setting the <strong>Hosting:Environment</strong> environment variable to <strong>Development</strong>, and restarting the application.
    </p>
```

rustd:
@rowanmiller are you going to update this PR or should we go ahead with my proposed change?

rowanmiller:
I can update it quickly with your suggestion... gimme a couple of min :smile:

rowanmiller:
Updated, ready for review

src/BaseTemplates/StarterWeb/Views/Shared/Error.cshtml
```diff
@@ -4,3 +4,11 @@
 
 <h1 class="text-danger">Error.</h1>
 <h2 class="text-danger">An error occurred while processing your request.</h2>
+
+<h3>Development Mode</h3>
+<p>
+    Swapping to <strong>Development</strong> environment will display more detailed information about the error that occurred.
+</p>
+<p>
+    Development environment can be enabled by setting the <strong>Hosting:Environment</strong> environment variable to <strong>Development</strong>, and restarting the application.
```
rustd: This name should match whatever is the latest in RC2. Looks good otherwise

rowanmiller: Do you know what the name is for RC2? I've been digging thru hosting code and it's a bit of a rabbit maze to work out exactly what env variable controls the name :smile:

rustd: ASPNETCORE_ENVIRONMENT from https://github.com/aspnet/Hosting/pull/640

rowanmiller: Thanks! Am I good to push this one with that change?

rustd: :shipit: We will update it later if there is a different name

blowdart:
At least point out that this should never be done on a live server due to security concerns.

rowanmiller:
@blowdart do you have some text we have used elsewhere? Or something you want us to use?

blowdart:
Well I'd rather you didn't add this at all. It's the same problem we had in ASP.NET 4. People would follow the advice blindly. It encourages poor practices.

rustd:
Hmm. Let's hold off on this for RC2 and see if we have a better solution in RTM.

rowanmiller:
Updated with some text about why not to do this... and the text is before the bit that tells you how to do it. I think we should push to get this into RC2 since it is a confusing issue that folks have hit and there is no way to discover this if you run outside of VS (other than reading docs).

rustd:
@blowdart ping. Is Rowan's proposal good enough for RC2 or is there a better solution here.

blowdart:
&lt;strong&gt; the should not and I'll acquiesce.

rowanmiller:
Done. @rustd will someone on your side merge?

rustd:
@phenning can merge.

PR #433, opened 2016-03-10
Update templates for changes in EF service registration

ajcvickers:
- No longer need to call AddEntityFramework and AddSqlServer. Only AddDbContext is needed.
- Need a constructor on the context that takes DbContextOptions
  - Previous "magic" that allowed the parameterless constructor has been removed
  - Identity repro will be updated with the new constructor patterns soon

@rowanmiller @divega

### DIFF
--- a/src/Rules/StarterWeb/IndividualAuth/Data/ApplicationDbContext.cs
+++ b/src/Rules/StarterWeb/IndividualAuth/Data/ApplicationDbContext.cs
@@ -10,6 +10,11 @@ namespace $safeprojectname$.Data
 {
     public class ApplicationDbContext : IdentityDbContext<ApplicationUser>
     {
+        public ApplicationDbContext(DbContextOptions<ApplicationDbContext> options)
+            : base(options)
+        {
+        }
+
         protected override void OnModelCreating(ModelBuilder builder)
         {
             base.OnModelCreating(builder);
--- a/src/Rules/StarterWeb/AI/IndividualAuth/Startup.cs
+++ b/src/Rules/StarterWeb/AI/IndividualAuth/Startup.cs
@@ -49,10 +49,8 @@ namespace $safeprojectname$
             // Add framework services.
             services.AddApplicationInsightsTelemetry(Configuration);
 
-            services.AddEntityFramework()
-                .AddSqlServer()
-                .AddDbContext<ApplicationDbContext>(options =>
-                    options.UseSqlServer(Configuration["Data:DefaultConnection:ConnectionString"]));
+            services.AddDbContext<ApplicationDbContext>(options =>
+                options.UseSqlServer(Configuration["Data:DefaultConnection:ConnectionString"]));
 
             services.AddIdentity<ApplicationUser, IdentityRole>()
                 .AddEntityFrameworkStores<ApplicationDbContext>()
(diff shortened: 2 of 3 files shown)


ajcvickers:
Not sure why the diff is all messed up. Diff shows fine in CodeFlow. I only changed:

``` C#
services.AddEntityFramework()
    .AddSqlServer()
    .AddDbContext<ApplicationDbContext>(options =>
        options.UseSqlServer(Configuration["Data:DefaultConnection:ConnectionString"]));
```

to

``` C#
services.AddDbContext<ApplicationDbContext>(options =>
    options.UseSqlServer(Configuration["Data:DefaultConnection:ConnectionString"]));
```

And added this constructor:

``` C#
public ApplicationDbContext(DbContextOptions<ApplicationDbContext> options)
    : base(options)
{
}
```

rustd:
Does this require a new build of EF runtime? When were the EF changes checked in? The templates use coherence-signed so these should be merged when there is a successful coherence-signed build

src/Rules/StarterWeb/IndividualAuth/Data/ApplicationDbContext.cs
```diff
@@ -1,21 +1,26 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Identity.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore;
+using $safeprojectname$.Models;
+
+namespace $safeprojectname$.Data
+{
+    public class ApplicationDbContext : IdentityDbContext<ApplicationUser>
+    {
+        public ApplicationDbContext(DbContextOptions<ApplicationDbContext> options)
```
rustd, line 13: Do we need to include this namespace?
using Microsoft.EntityFrameworkCore.Infrastructure;

ajcvickers:
Yes, it requires a new build. I'm not going to merge this yet. @rowanmiller asked me to make the PR now so the changes could be seen in the template review meeting this afternoon.

src/Rules/StarterWeb/IndividualAuth/Data/ApplicationDbContext.cs
ajcvickers, line 13: Yes it does right now, but I think we should move DbContextOptions instead of adding the namespace. @rowanmiller @divega

rustd:
Thanks!

src/Rules/StarterWeb/IndividualAuth/Data/ApplicationDbContext.cs
rowanmiller, line 13: :+1:

phenning:
@ajcvickers It looks like we need this now with the latest coherence-signed?

[09:54:57] :     [Step 2/4] D:\b\w\5f4bbb4d57087cbf\artifacts\build\Test\StarterWeb.IndividualAuth\Startup.cs(46,22): error CS1061: 'IServiceCollection' does not contain a definition for 'AddEntityFramework' and no extension method 'AddEntityFramework' accepting a first argument of type 'IServiceCollection' could be found (are you missing a using directive or an assembly reference?) [D:\b\w\5f4bbb4d57087cbf\test\VerifyTemplates.csproj]

phenning:
Verified that taking this fixes the template build with the latest coherence signed, will merge now.

PR #434, opened 2016-03-10
Remove web command profile from WebAPI launchSetting.json

### LINKED ISSUE #432: Remove web command profile from WebAPI launchSetting.json
The profile is no longer created in the tooling, so it should be removed

phenning:
Fixes #432 

@rustd 
@BillHiebert 

Since we no longer have a web profile by default, we should remove this from the template.

### DIFF
--- a/src/Rules/WebAPI/WindowsAuth/Properties/launchSettings.json
+++ b/src/Rules/WebAPI/WindowsAuth/Properties/launchSettings.json
@@ -11,14 +11,6 @@
       "environmentVariables" : {
         "ASPNET_ENVIRONMENT": "Development"
       }
-    },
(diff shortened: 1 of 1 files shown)


rustd:
Probably. It was brought up in the templates review that Bill was working on the replacement for this. /cc @sayedihashimi

phenning:
We should still remove it until we have the replacement, since we only have it for WebAPI Windows auth only (not there for non-auth)

rustd:
Yes, we would remove it and replace it with what Bill is working on

phenning:
OK, I'll merge this change and add the new version back in a later commit, after that work is done.

PR #449, opened 2016-03-17
Rename "ASP.NET Core 1" references to "ASP.NET Core"

phenning:
Fix the branding. Also changing this in tooling as well, sent out separate review for that.

@joeloff @rustd


src/BaseTemplates/WebAPI.vstemplate
```diff
@@ -1,8 +1,8 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <VSTemplate Version="3.0.0" Type="Project" xmlns="http://schemas.microsoft.com/developer/vstemplate/2005" xmlns:sdk="http://schemas.microsoft.com/developer/vstemplate-sdkextension/2010">
   <TemplateData>
-    <Name>ASP.NET Core 1 Preview Web API</Name>
-    <Description>A project for creating an ASP.NET Core 1 Preview Web API</Description>
+    <Name>ASP.NET Core Preview Web API</Name>
```
joeloff: Why do these say Preview and the other don't?

phenning: @joeloff It looks like all the web vstemplates say preview. I'm not sure what you mean here...

PR #455, opened 2016-03-23
Remove unused namespaces

phenning:
Removing unused namespaces in templates
Also, there was some that were mis-ordered, and in another case a missing using in one of the AI templates.

@rustd


rustd:
:shipit:

phenning:
Merged

PR #466, opened 2016-03-27
:bug: Remove extra dots from page title. Fixes #465

### LINKED ISSUE #465: Extra dot to remove
`ForgotPassword` template:
https://github.com/aspnet/Templates/blob/7b83d3628ed40a2637fe36946ba8185ac232a190/src/Rules/StarterWeb/IndividualAuth/Views/Account/ForgotPassword.cshtml
The title ends with dot:
https://github.com/aspnet/Templates/blob/7b83d3628ed40a2637fe36946ba8185ac232a190/src/Rules/StarterWeb/IndividualAuth/Views/Account/ForgotPassword.cshtml#L6
where `Title` variable end with question mark.
So template renders:

```
Forgot your password?.
```

peterblazejewicz:
An itsy-bitsy one
The defualt title ends with a question mark, so by
default user is presented with the extra dot:

```
Forgot your password?.
```

Thanks!

### DIFF
--- a/src/Rules/StarterWeb/IndividualAuth/Views/Account/ForgotPassword.cshtml
+++ b/src/Rules/StarterWeb/IndividualAuth/Views/Account/ForgotPassword.cshtml
@@ -3,7 +3,7 @@
     ViewData["Title"] = "Forgot your password?";
 }
 
-<h2>@ViewData["Title"].</h2>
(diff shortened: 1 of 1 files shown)


rustd:
:shipit: @phenning this should be good to me merged.

PR #467, opened 2016-03-27
:art: Use string interpolation when building email links

peterblazejewicz:
This commit simplifies email links generation by using
C# 6 string interpolation feature.
The quotes in HTML5 are double, single or no quotes.
This commit uses single quotes to not use escape chars.

Thanks!


rustd:
:shipit: @phenning this should be good to be merged. /cc @sayedihashimi

PR #477, opened 2016-03-31
reloadOnChange cleanup

HaoK:
cc @divega @rustd @phenning

### DIFF
--- a/src/BaseTemplates/StarterWeb/Startup.cs
+++ b/src/BaseTemplates/StarterWeb/Startup.cs
@@ -19,10 +19,14 @@ namespace $safeprojectname$
             // Set up configuration sources.
             var builder = new ConfigurationBuilder()
                 .SetBasePath(appEnv.ApplicationBasePath)
-                .AddJsonFile("appsettings.json")
+                .AddJsonFile(source =>
+                {
+                    source.Path = "appsettings.json";
+                    source.ReloadOnChange = true;
+                })
                 .AddJsonFile($"appsettings.{env.EnvironmentName}.json", optional: true)
                 .AddEnvironmentVariables();
-            Configuration = builder.Build().ReloadOnChanged("appsettings.json");
+            Configuration = builder.Build();
         }
 
         public IConfigurationRoot Configuration { get; set; }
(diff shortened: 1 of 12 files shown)


rustd:
:shipit: once we have a co-herence signed build.

src/BaseTemplates/StarterWeb/Startup.cs
```diff
@@ -19,10 +19,14 @@ public Startup(IHostingEnvironment env, IApplicationEnvironment appEnv)
             // Set up configuration sources.
             var builder = new ConfigurationBuilder()
                 .SetBasePath(appEnv.ApplicationBasePath)
-                .AddJsonFile("appsettings.json")
+                .AddJsonFile(source =>
+                {
+                    source.Path = "appsettings.json";
+                    source.ReloadOnChange = true;
```
divega, line 25: Didn't we create a simple overload for this? Does the simplest/most common usage in our default template require a lambda for this provider while it doesn't for the others? :disappointed:

rustd, line 25: AddJsonFile("appsettings.json", reloadonChange=true) would be better here. no?

rustd:
@HaoK @divega @shirhatti pinging on this. Are we making changes to config APIs to improve this for RC2?

src/BaseTemplates/StarterWeb/Startup.cs
divega, line 25: I now remember there was a big discussion around this and ended up keeping the existing overloads but adding the lambda syntax to support the new parameters (including reload-on-change). Given that I believe we shouldn't try to rush any more changes in this area and that this is the best we can do for RC2. But we should take another look early in the next milestone and if possible cleanup and rationalize the overloads to have something that feels more cohesive. I have filed https://github.com/aspnet/Configuration/issues/420.

divega:
@rustd this is ready as far as I am concerned. Should @HaoK push it or should we wait for someone working on templates to do it?

rustd:
@phenning can you please merge this once there is a successful coherence signed builds with these changes.

PR #482, opened 2016-04-06
Update jquery-validation-unobtrusive to use 3.2.6 version

nithyachalla:
@rustd @phenning @Eilon


src/BaseTemplates/StarterWeb/wwwroot/lib/jquery-validation-unobtrusive/.bower.json
```diff
@@ -1,6 +1,6 @@
 {
```
nithyachalla: I am not clear why this file is checked in...this should be auto generated when bower installs the packages?

PR #485, opened 2016-04-07
SetBasePath -> SetProjectPath

phenning:
@HaoK @rustd @BillHiebert

### DIFF
--- a/src/Rules/StarterWeb/AI/IndividualAuth/Startup.cs
+++ b/src/Rules/StarterWeb/AI/IndividualAuth/Startup.cs
@@ -34,7 +34,7 @@ namespace $safeprojectname$
             if (env.IsDevelopment())
             {
                 // For more details on using the user secret store see http://go.microsoft.com/fwlink/?LinkID=532709
-                builder.AddUserSecrets();
+                builder.SetProjectPath(env.ContentRootPath).AddUserSecrets();
 
                 // This will push telemetry data through Application Insights pipeline faster, allowing you to view results immediately.
                 builder.AddApplicationInsightsSettings(developerMode: true);
--- a/src/Rules/StarterWeb/AI/OrganizationalAuth/Multiple/Startup.cs
+++ b/src/Rules/StarterWeb/AI/OrganizationalAuth/Multiple/Startup.cs
@@ -32,7 +32,7 @@ namespace $safeprojectname$
             if (env.IsDevelopment())
             {
                 // For more details on using the user secret store see http://go.microsoft.com/fwlink/?LinkID=532709
-                builder.AddUserSecrets();
+                builder.SetProjectPath(env.ContentRootPath).AddUserSecrets();
 
                 builder.AddApplicationInsightsSettings(developerMode: true);
             }
--- a/src/Rules/StarterWeb/AI/OrganizationalAuth/Single/Startup.cs
+++ b/src/Rules/StarterWeb/AI/OrganizationalAuth/Single/Startup.cs
@@ -30,7 +30,7 @@ namespace $safeprojectname$
             if (env.IsDevelopment())
             {
                 // For more details on using the user secret store see http://go.microsoft.com/fwlink/?LinkID=532709
-                builder.AddUserSecrets();
+                builder.SetProjectPath(env.ContentRootPath).AddUserSecrets();
 
                 builder.AddApplicationInsightsSettings(developerMode: true);
             }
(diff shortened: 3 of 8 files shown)


phenning:
Reacting to 
https://github.com/aspnet/UserSecrets/commit/41e3e6d07cab1e51733e6143a5753e30b740625c

phenning:
@DamianEdwards

phenning:
for templates that don't have usersecrets, we totally remove the call.

rustd:
Why would we remove it? Leaving it there would be better only.

phenning:
@rustd SetProjectPath resides in the UserSecrets package, for templates that don't use UserSecrets, we need to remove it. Are you suggesting that we leave the call to SetBasePath in there for those templates? Should we just leave SetBasePath in there all the time, and just add the call to SetProjectPath for the templates that use UserSecrets?

phenning:
I'd make the argument that SetProjectPath maybe should be in a different package so we can safely call it from any template...

HaoK:
As it stands right now SetProjectPath is only used by user secrets. SetBasePath is for the file based config providers. They are no longer related. We might want to revisit this again @divega post rc2

divega:
I thought the idea was that `SetProjectPath()` was only going to be necessary in rare testing scenarios because if not provided [we default to `AppContext.BaseDirectory`](https://github.com/aspnet/UserSecrets/blob/dev/src/Microsoft.Extensions.Configuration.UserSecrets/ConfigurationExtensions.cs#L21), but if `env.ContentRootPath` needs to be passed I believe a regular parameter on `AddUserSecrets()` works much better. 

Then the challenge is that we already have an overload that takes a string: `configuration.AddUserSecrets(userSecretsId)`. Maybe we should name the other one `AddUserSecretsFromProject()` instead? cc @HaoK

divega:
@glennc to confirm whether `AppContext.BaseDirectory` should be enough or otherwise if we should make passing the project directory a first class parameter on the add method.

divega:
BTW, since we are making a breaking change here already I think it is worth trying to avoid making a new one in the next milestone :smile:

rustd:
@divega can you please open an RC2 bug to track this then

divega:
I had a discussion with @DamianEdwards @glennc and @davidfowl . Turns out **we should not remove the calls to `SetBasePath()`** nor replace them with calls to `SetProjectPath()`. The latter is going away shortly.

So @phenning please do not push this change.

phenning:
We need a short term solution to unblock WTE with the current bits that we have.

instead of the current change, how about we just change:

`builder.AddUserSecrets()`

to 

`builder.SetProjectPath(env.ContentRootPath).AddUserSecrets() in the templates that need them.`

If we do not do at least that, default F5 experience in VS is broken.

divega:
@phenning I think that is reasonable until https://github.com/aspnet/UserSecrets/issues/70 is fixed.

phenning:
Updated PR with that. We can revert this after aspnet/UserSecrets#70  is fixed

divega:
:shipit: LGTM, not sure if you need another review.

src/Rules/StarterWeb/AI/IndividualAuth/Startup.cs
```diff
@@ -34,7 +34,7 @@ public Startup(IHostingEnvironment env)
             if (env.IsDevelopment())
             {
                 // For more details on using the user secret store see http://go.microsoft.com/fwlink/?LinkID=532709
-                builder.AddUserSecrets();
+                builder..SetProjectPath(env.ContentRootPath).AddUserSecrets();
```
smitpatel: is it really `..`?

phenning: Will fix before merging.
- Phil

On Apr 7, 2016, at 5:33 PM, Smit Patel <notifications@github.com<mailto:notifications@github.com>> wrote:

In src/Rules/StarterWeb/AI/IndividualAuth/Startup.cshttps://github.com/aspnet/Templates/pull/485#discussion_r58966757:

> @@ -34,7 +34,7 @@ public Startup(IHostingEnvironment env)
>              if (env.IsDevelopment())
>              {
>                  // For more details on using the user secret store see http://go.microsoft.com/fwlink/?LinkID=532709
> -                builder.AddUserSecrets();
> -                builder..SetProjectPath(env.ContentRootPath).AddUserSecrets();

is it really ..?

##

PR #496, opened 2016-04-11
Fix up formatting in _Layout.cshtml files and ManageLogin.cshtml

phenning:
@vkichline @rustd @sayedihashimi @balachir 

Fixing some incorrectly formatted files


rustd:
:shipit:

PR #497, opened 2016-04-12
Move web.config

JunTaoLuo:
We now expect to have the `web.config` file to be located in the application root instead of `wwwroot`. We also made some changes to the default visible attributes.

cc @Tratcher @shirhatti @phenning

### DIFF
--- a/src/BaseTemplates/WebAPI/project.json
+++ b/src/BaseTemplates/WebAPI/project.json
@@ -41,7 +41,8 @@ $endif$
   "content": [
     "wwwroot",
     "Views",
-    "appsettings.json"
+    "appsettings.json",
+    "web.config"
   ],
 
   "exclude": [
--- a/src/BaseTemplates/WebAPI/web.config
+++ b/src/BaseTemplates/WebAPI/web.config
@@ -9,6 +9,6 @@
     <handlers>
       <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModule" resourceType="Unspecified"/>
     </handlers>
-    <aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS%" startupTimeLimit="3600" stdoutLogEnabled="false"/>
+    <aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS%" stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" forwardWindowsAuthToken="false"/>
   </system.webServer>
 </configuration>
--- a/src/BaseTemplates/EmptyWeb.vstemplate
+++ b/src/BaseTemplates/EmptyWeb.vstemplate
@@ -22,9 +22,8 @@
       <ProjectItem ReplaceParameters="true">project.json</ProjectItem>
       <ProjectItem ReplaceParameters="true" OpenInEditor="true">Startup.cs</ProjectItem>
       <ProjectItem ReplaceParameters="true" TargetFileName="Project_Readme.html" OpenInWebBrowser="true">Project_Readme.html</ProjectItem>
-      <Folder Name="wwwroot" TargetFolderName="wwwroot">
-        <ProjectItem ReplaceParameters="true" TargetFileName="web.config">web.config</ProjectItem>
-      </Folder>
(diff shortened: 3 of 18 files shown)


phenning:
//cc @BillHiebert any tooling impact for the web.config attribute changes?

src/BaseTemplates/WebAPI/web.config
```diff
@@ -9,6 +9,6 @@
     <handlers>
       <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModule" resourceType="Unspecified"/>
     </handlers>
-    <aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS%" startupTimeLimit="3600" stdoutLogEnabled="false"/>
+    <aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS%" stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" forwardWindowsAuthToken="false"/>
```
phenning, line 12: Does forwardWindowsAuthToken need to be set to true for the Windows Auth template? If so we need to lay down a different web.config for that template

src/BaseTemplates/WebAPI/project.json
```diff
@@ -41,7 +41,8 @@ $endif$
   "content": [
     "wwwroot",
     "Views",
-    "appsettings.json"
+    "appsettings.json",
+    "web.config"
```
phenning, line 45: For the project.json changes, you also need to update all the project.json files under src\Rules**\project.json

src/BaseTemplates/WebAPI/web.config
phenning, line 12: Also, you need to update the vstemplate files in under src\BaseTemplates to move the web.config out of the wwwroot folder and into the root folder. (https://github.com/aspnet/Templates/blob/dev/src/BaseTemplates/EmptyWeb.vstemplate#L26) I'd leave the empty folder element so the folder gets created when the template is instantiated

phenning:
I had another PR pending for just moving the files, but was missing the project.json changes. I'm going to close my PR and we can continue to move forward with the change in this PR.

src/BaseTemplates/WebAPI/web.config
phenning, line 12: I just talked to Bill about the forwardWindowsAuthToken, we should probably have a different web.config that is explicit, but it isn't needed for RC2 since the tooling will make sure it has the right value.

rustd, line 12: Can we please open a bug to track this?

src/BaseTemplates/WebAPI/project.json
JunTaoLuo, line 45: Got it thanks!

src/BaseTemplates/WebAPI/web.config
JunTaoLuo, line 12: Not sure if this is a problem but empty directories is not picked up by git. So even if the user generates a template with the empty wwwroot folder, if they clone it somewhere else, they will be missing that directory.

JunTaoLuo, line 12: Also should I remove `forwardWindowsAuthToken="false"` attribute from WebAPI template only?

phenning, line 12: @JunTaoLuo that's fine, realistically the user will add _something_, if not its probably ok.

phenning, line 12: @JunTaoLuo I think we can keep it there for the WebAPI template, no?

JunTaoLuo, line 12: @phenning Then I think this PR is ready to go?

phenning:
Changes look good, but can we hold it until @BillHiebert is ready to go with the needed tooling changes so we can still take other template changes unrelated to this change in the meanwhile?

JunTaoLuo:
@phenning of course.

phenning:
This should be ok to merge now, merging it.

JunTaoLuo:
Thanks, this needs to be merged to dev too right?

phenning:
Yup, I’ll merge it over by EOD with any other changes.

PR #506, opened 2016-04-15
Add missing tools node to WebAPI project.json

JunTaoLuo:
@phenning I think these `project.json` are missing the `"Microsoft.AspNetCore.Server.IISIntegration.Tools"` dependencies even though they are using `dotnet publish-iis`. If you agree, we should seek rc2 approval for these updates.


phenning:
:shipit:

phenning:
I double checked and it doesn't look like we are missing this anywhere else.

phenning:
@rustd ok to merge?

rustd:
yup

phenning:
merged

PR #514, opened 2016-04-19
Add Server GC to project.json and reorder sections

### LINKED ISSUE #510: Use Server GC
Is there any reason not to turn on server GC in our templates? This is possible now with dotnet so we should do it :+1: 

/cc @DamianEdwards @Eilon

phenning:
Fixes #510 

@rynowak @rustd @DamianEdwards


rustd:
:shipit:

PR #521, opened 2016-04-22
Include self-host launch settings by default

### LINKED ISSUE #519: Include self-host launch settings by default
The templates currently include a profile for launching IIS Express, but nothing for self-host. Self-host should be included by default.

Something like this in the launchSettings.json:

```
    "self-host": {
      "commandName": "Project",
      "launchBrowser": true,
      "launchUrl": "http://localhost:5000/",
      "environmentVariables": {
        "ASPNETCORE_ENVIRONMENT": "Development"
      }
    }
```

phenning:
Fixes #519

//cc @rustd @Tratcher


rustd:
:shipit:

PR #522, opened 2016-04-25
Update OrgAuth StarterWeb for runtime changes

phenning:
@Tratcher @rustd 

Diff is a bit odd due to reorganization of the directory structure for templates to accommodate the AD read version of Startup.cs  (The NoRead one is differed from the original version, the Read one is being displayed as a new file)


src/Rules/StarterWeb/OrganizationalAuth/Multiple/Common/Controllers/AccountController.cs
```diff
@@ -17,13 +17,11 @@ public IActionResult SignIn()
                 new AuthenticationProperties { RedirectUri = "/" }, OpenIdConnectDefaults.AuthenticationScheme);
         }
 
-        public async Task<IActionResult> SignOut()
+        public IActionResult SignOut()
```
rustd, line 20: Can you please make these changes for MultIOrg templates as well?

src/Rules/StarterWeb/OrganizationalAuth/Single/Common/Controllers/AccountController.cs
```diff
@@ -17,13 +17,11 @@ public IActionResult SignIn()
                 new AuthenticationProperties { RedirectUri = "/" }, OpenIdConnectDefaults.AuthenticationScheme);
         }
 
-        public async Task<IActionResult> SignOut()
```
rustd, line 20: Can you please make this change for MultiOrg template as well?

src/Rules/WebAPI/OrganizationalAuth/Single/Startup.cs
```diff
@@ -52,7 +52,7 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerF
             app.UseJwtBearerAuthentication(new JwtBearerOptions
             {
                 Authority = Configuration["Authentication:AzureAd:AADInstance"] + Configuration["Authentication:AzureAd:TenantId"],
-                Audience = Configuration["Authentication:AzureAd:AudienceUrl"]
```
rustd, line 55: I am assuming there will be a Org Tooling/ publish update for these settings.

rustd, line 55: For Web API, this covers both Read and NonRead cases?

src/Rules/StarterWeb/OrganizationalAuth/Multiple/Common/Controllers/AccountController.cs
phenning, line 20: Already done, this is the MultiOrg. :)

src/Rules/StarterWeb/OrganizationalAuth/Single/Common/Controllers/AccountController.cs
phenning, line 20: Done, see above

src/Rules/WebAPI/OrganizationalAuth/Single/Startup.cs
phenning, line 55: Correct, there will a corresponding Tooling update here as well.

phenning, line 55: Yes, this should cover both read and non read, as currently for web api they are identical

rustd:
:shipit:

phenning:
Merged

PR #524, opened 2016-04-25
Remove duplicate .ReloadOnChanged("appsettings.json") in Startup.cs

### LINKED ISSUE #523: Remove duplicate .ReloadOnChanged("appsettings.json") in Startup.cs
This is redundant with the earlier setting of ReloadOnChanged for the source, and will be removed soon.  All Startups should have this call removed so we can remove the API from configuration

cc @divega

phenning:
Fixes #523 

//cc @HaoK , @rustd


rustd:
:shipit:

PR #536, opened 2016-05-03
React to project.json schema changes and tooling package renames

phenning:
//cc @DamianEdwards @rustd @anurse @moozzyk @prafullbhosale @NTaylorMullen

### DIFF
--- a/src/BaseTemplates/EmptyWeb/project.json
+++ b/src/BaseTemplates/EmptyWeb/project.json
@@ -39,8 +39,7 @@ $endif$
     "include": [
       "wwwroot",
       "web.config"
-    ],
-    "exclude": [ "node_modules" ]
+    ]
   },
 
   "scripts": {$if$ ($aspnet_useplatformhandler$ == false)$else$
--- a/src/BaseTemplates/EmptyWeb/project.json
+++ b/src/BaseTemplates/EmptyWeb/project.json
@@ -26,7 +26,7 @@
     "$netframeworkversion$": { }$endif$
   },
 
-  "compilationOptions": {
+  "buildOptions": {
     "emitEntryPoint": true,
     "preserveCompilationContext": true
   },
--- a/src/BaseTemplates/EmptyWeb/project.json
+++ b/src/BaseTemplates/EmptyWeb/project.json
@@ -32,19 +32,16 @@
   },
 $if$ ($context$ == WebCore)
   "runtimeOptions": {
-    "configProperties": {
-      "System.GC.Server": true
-    }
+    "gcServer": true
   },
 $endif$
-  "content": [
-    "wwwroot",
-    "web.config"
-  ],
-
-  "exclude": [
-    "node_modules"
-  ],
+  "publishOptions": {
+    "include": [
+      "wwwroot",
+      "web.config"
+    ],
+    "exclude": [ "node_modules" ]
+  },
 
   "scripts": {$if$ ($aspnet_useplatformhandler$ == false)$else$
     "postpublish": [ "dotnet publish-iis --publish-folder %publish:OutputPath% --framework %publish:FullTargetFramework%" ]$endif$
--- a/src/BaseTemplates/StarterWeb/project.json
+++ b/src/BaseTemplates/StarterWeb/project.json
@@ -43,7 +43,7 @@
     "$netframeworkversion$": { }$endif$
   },
 
-  "compilationOptions": {
+  "buildOptions": {
     "emitEntryPoint": true,
     "preserveCompilationContext": true
   },
--- a/src/BaseTemplates/StarterWeb/project.json
+++ b/src/BaseTemplates/StarterWeb/project.json
@@ -4,12 +4,12 @@
       "version": "__NetCorePlatformVersion__",
       "type": "platform"
     },$endif$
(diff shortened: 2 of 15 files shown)


src/BaseTemplates/EmptyWeb/project.json
```diff
@@ -26,25 +26,22 @@
...
-
-  "exclude": [
-    "node_modules"
-  ],
+  "publishOptions": {
+    "include": [
+      "wwwroot",
+      "web.config"
+    ],
+    "exclude": [ "node_modules" ]
```
DamianEdwards: The `"exclude"` shouldn't be needed at all. You will however need to include the `appsettings.json`

DamianEdwards: And you need to include "Views" for MVC apps too

DamianEdwards: Just saw these are separate files, carry on.

src/BaseTemplates/StarterWeb/project.json
```diff
@@ -43,27 +43,24 @@
...
-    "node_modules"
-  ],
+  "publishOptions": {
+    "include": [
+      "wwwroot",
+      "Views",
+      "appsettings.json",
+      "web.config"
+    ],
+    "exclude": [ "node_modules" ]
```
DamianEdwards: Shouldn't need this `"exclude"` as above

src/BaseTemplates/WebAPI/project.json
```diff
@@ -33,27 +33,24 @@
...
-    "node_modules"
-  ],
+  "publishOptions": {
+    "include": [
+      "wwwroot",
+      "Views",
+      "appsettings.json",
+      "web.config"
+    ],
+    "exclude": [ "node_modules" ]
```
DamianEdwards: Shouldn't need this

src/Rules/StarterWeb/AI/IndividualAuth/project.json
```diff
@@ -61,6 +58,13 @@
         "portable-net45+win8"
       ]
     },
+    "Microsoft.Extensions.CodeGeneration.Tools": {
+      "version": "__AspNetCorePlatformVersion__",
+      "imports": [
+        "portable-net45+win8+dnxcore50",
```
DamianEdwards, line 64: Do we know why **both** of these imports are required here?

src/Rules/StarterWeb/AI/IndividualAuth/project.json
```diff
@@ -78,27 +82,24 @@
...
-    "node_modules"
-  ],
+  "publishOptions": {
+    "include": [
+      "wwwroot",
+      "Views",
+      "appsettings.json",
+      "web.config"
+    ],
+    "exclude": [ "node_modules" ]
```
DamianEdwards: Shouldn't need this

src/Rules/StarterWeb/AI/NoAuth/project.json
```diff
@@ -44,27 +44,24 @@
...
-    "node_modules"
-  ],
+  "publishOptions": {
+    "include": [
+      "wwwroot",
+      "Views",
+      "appsettings.json",
+      "web.config"
+    ],
+    "exclude": [ "node_modules" ]
```
DamianEdwards: Shouldn't need this

src/BaseTemplates/WebAPI/project.json
phenning: I think the reason we had the exludes in there for all the web templates is so if customers start installing node packages we automatically exclude them, even though these templates don't ship with node modules by default.

src/Rules/StarterWeb/AI/IndividualAuth/project.json
phenning, line 64: Yes, net45+win8+dnxcore50 is required to get the proper version of json.net, just specifying net45+win8 nets you a json.net that has a fx dependency on System.Runtime 2.0.5.0

net45+win8 is required for Ix.Async and Remotion

src/Rules/StarterWeb/AI/IndividualAuth/project.json
phenning: Don't we actually need it for all StarterWebs (assuming that its purpose is what I think it is)

NTaylorMullen:
Razor tooling bits look good. :+1:

prafullbhosale:
Scaffolding bits look good too!

src/BaseTemplates/WebAPI/project.json
rustd: yeah. that's what I remember @madskristensen do we need to exclude the node_modules?

madskristensen: If the template doesn't have a node_modules folder then there is no reason to exclude it at all. That would just be confusing

rustd: Of course. @phenning we only need it only for the Starter Web templates then which have a node_modules folder

phenning: We don't need it at all anymore, due to the way concerns will split between publish and build. I confirmed this offline with Damian.

PR #555, opened 2016-05-09
Remove OptionsModel using from AI _ViewImports.cshtml

phenning:
@rustd


rustd:
:shipit: please see if there are other places AI is called which has this code.

phenning:
merged

PR #557, opened 2016-05-10
Update rendering of the "name" field in bower.json files

### LINKED ISSUE #550: Update rendering of the "name" field in bower.json files
Updated Bower renders now warnings when a name field is not _safe_ name, including lower case enforcement. This was wrongly implemented in Bower < 1.7.9 as an error, not as the warning, so users could hit a problem.
I think the VS templates use a hardcoded key(?), which will cause a warning in Bower 1.7.9:

> bower                     invalid-meta The "name" is recommended to be lowercase, can contain digits, dots, dashes

https://github.com/aspnet/Templates/blob/dev/src/BaseTemplates/StarterWeb/bower.json#L2
https://github.com/bower/bower/issues/2233

Thanks!

phenning:
Fixes #550 
//cc: @rustd, @madskristensen

### DIFF
--- a/src/BaseTemplates/StarterWeb/bower.json
+++ b/src/BaseTemplates/StarterWeb/bower.json
@@ -1,5 +1,5 @@
 {
-  "name": "ASP.NET",
+  "name": "asp.net",
   "private": true,
   "dependencies": {
     "bootstrap": "3.3.6",


madskristensen:
LGTM

rustd:
:shipit:

phenning:
Merged

Issue #591, opened 2016-06-06
Include asp-area="" in links under Views/Shared/ views


javiercn:
- It teaches the customers what the right thing to do when they share views.
- It prevents issues that happen when a customer decides to include an area and their layout links suddenly change depending on whether they are rendering a view within an area or not.
- It will prevent a flood of incoming issues like https://github.com/aspnet/Mvc/issues/4815

phenning:
Thoughts? Preview2 tooling or post?

//cc @mlorbetske @sayedihashimi @Eilon

Eilon:
I'm not too concerned about adding this at all. Most sites don't use areas, so I don't really want to add extra stuff to the template in case someone adds areas. There's lots of other stuff we also don't put in the templates that you might need if you then make other changes to the template.

phenning:
This one seems rather ugly though for customer to fix, and its not obvious what you need to do in order to fix it.

Eilon:
Ok, thought about it some more, would be nice to add this. But perhaps not super urgent, so if it's too busy for this release, it can wait.

phenning:
Thanks Eilon, we'll be able to get this in along with #595 easily.

PR #593, opened 2016-06-07
Improve Area support in default templates

phenning:
@javiercn @mlorbetske @sayedihashimi @joeloff

### DIFF
--- a/src/BaseTemplates/StarterWeb/project.json
+++ b/src/BaseTemplates/StarterWeb/project.json
@@ -71,6 +71,7 @@ $endif$
     "include": [
       "wwwroot",
       "Views",
+      "Areas\**\Views",
       "appsettings.json",
       "web.config"
     ]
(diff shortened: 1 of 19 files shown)


phenning:
Fixes #591

joeloff:
:shipit:

phenning:
@rynowak, added the publish options change to this now as well

src/BaseTemplates/StarterWeb/project.json
```diff
@@ -71,6 +71,7 @@ $endif$
     "include": [
       "wwwroot",
       "Views",
+      "Areas\**\Views",
```
rynowak, line 74: Does this do the right thing? Just making sure you double checked that this is actually the right include

phenning, line 74: Yup this works, verified in a sample app

phenning, line 74: Actually the include should be Areas/**/Views / instead of \

PR #599, opened 2016-06-09
Remove comment about nuget packages

rchande:
@phenning Can you merge/let me know if this is the right branch?
cc @sayedihashimi @Pilchie


phenning:
Correct branch, merging!

Pilchie:
This needs to be inserted somewhere, right?  @phenning - will that be taken care of automatically, or does @rchande need to do it?

phenning:
Nope, we'll take care of inserting the package as soon as we get a template build. It should go in today.

Pilchie:
Thanks!

PR #603, opened 2016-06-13
Remove exclusions of node_modules folder from compilation

scottaddie:
The `project.json` files for the various project templates are currently excluding npm's `node_modules` folder during compilation. This seems unnecessary. There will no longer be a `node_modules` folder by default, since npm packages are no longer included in the templates. For example, Gulp and all of its plugins were removed.

### DIFF
--- a/src/BaseTemplates/EmptyWeb/project.json
+++ b/src/BaseTemplates/EmptyWeb/project.json
@@ -28,12 +28,7 @@
 
   "buildOptions": {
     "emitEntryPoint": true,
-    "preserveCompilationContext": true,
-    "compile": {
-      "exclude": [
-        "node_modules"
-      ]
-    }
+    "preserveCompilationContext": true
   },
 $if$ ($context$ == WebCore)
   "runtimeOptions": {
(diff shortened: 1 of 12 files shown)


phenning:
We may actually want to keep this include here by default for customers who explicitly start using any node modules in their projects.

peterblazejewicz:
> We may actually want to keep this include here by default for customers who explicitly start using any node modules in their projects.

I've hit that problem. The Angular 2 projects NPM e2e tests contains selenium sources with C# code, so the `exclude` was required.
https://github.com/peterblazejewicz/polymer-dotnet-recipes/commit/8d09499a415c555950135765032340b680699866

But I'm not sure if `node_modules` should be left in templates for this reason.

phenning:
@mlorbetske, @madskristensen @sayedihashimi , thoughts?

madskristensen:
The interesting thing is that npm package can now include C# code for compilation into the app, which is very cool. It could also be exploited, but so could nuget packages from random sources. I'm fine with removing the exclude from project.json

scottaddie:
> We may actually want to keep this include here by default for customers who explicitly start using any node modules in their projects.

My stance is that customers should be expected to manually add this exclude if they begin using npm packages. There are many ASP.NET devs in the enterprise who would rather not learn and use npm unless absolutely necessary. For that crowd, the `node_modules` exclude here is nothing more than cruft.

mlorbetske:
lgtm

PR #606, opened 2016-06-14
Individual Auth ManageController should guard against null current user.

### LINKED ISSUE #596: Individual Auth ManageController should guard against null current user.
The various actions don't handle the situation where GetCurrentUser returns null very gracefully, we should add a check and show an error or just redirect back to Home or something in this situation instead.

phenning:
Fixes #596

@HaoK


HaoK:
:shipit:

PR #607, opened 2016-06-15
Replace magic action name strings with nameof operator

scottaddie:
This PR replaces magic strings being passed as the action name parameters in the `Url.Action` calls. This will help ensure that as controller action methods are renamed, the `Url.Action` calls aren't forgotten. It's also consistent with what's already being done in the `RedirectToAction` calls.


peterblazejewicz:
These are not magic strings IMO. The proposed code is just more technical solution I think, but is not used in documentation or example content:
https://github.com/aspnet/MusicStore/blob/19325a6ccecaf24122d19a89502b9fd2342864ac/src/MusicStore/Controllers/AccountController.cs#L163
There are other ways to avoid string literals (we use global store pattern with constants for controllers and actions at work)

phenning:
Moved to post preview2.

PR #624, opened 2016-06-20
Edited the GetCurrentUserAsync function in the controllers.

adamradocz:
- Added the async  and the await tag.


phenning:
No real functional reason for doing this. You need to await in the calling method anyway.

PR #628, opened 2016-06-22
move dotnet bundle from precompile to prepublish to fix issues with i…

phenning:
…ncremental builds in VS.

@mlorbetske @barrytang @BillHiebert @balachir

### DIFF
--- a/src/BaseTemplates/StarterWeb/project.json
+++ b/src/BaseTemplates/StarterWeb/project.json
@@ -60,8 +60,7 @@ $endif$
   },
 
   "scripts": {
-    "precompile": [ "dotnet bundle" ],
-    "prepublish": [ "bower install" ]$if$ ($aspnet_useplatformhandler$ == false)$else$,
+    "prepublish": [ "bower install", "dotnet bundle" ]$if$ ($aspnet_useplatformhandler$ == false)$else$,
     "postpublish": [ "dotnet publish-iis --publish-folder %publish:OutputPath% --framework %publish:FullTargetFramework%" ]$endif$
   }
 }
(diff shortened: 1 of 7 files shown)


phenning:
Merging now to get the build-chain rolling

peterblazejewicz:
This assumes that users do not introduce any change to client side code before publish? What about development? What is a workflow for CSS and JS changes before publish? Will changes be picked by bundler when developing project, but before publish?
Thanks!

phenning:
@peterblazejewicz 

The issue we are mitigating here is this:

When CTRL-F5 the project in Visual Studio, then attempting to run scaffolding or other EF commands from the command line, results in an attempted build.

This build then fails, causing the scaffold or command to fail due to the file being in use. By moving this to prepublish we still cause the minification on publish to the production environment and fix the running/scaffold workflow.

In Visual Studio, the recommended solution is to install the VS Extension in order to run the task via Task runner explorer before publish, for command line development, you can still run "dotnet bundle".

phenning:
@madskristensen FYI.

peterblazejewicz:
Thanks!. Outside of VS/VSCode context - is it safe to advice users to `dotnet bundle watch` (recently added)?
I know Mads input and Wiki on this subject: https://github.com/madskristensen/BundlerMinifier/wiki
but I never used watch with Bundler yet.

mlorbetske:
@peterblazejewicz yep, there's no VS/VSCode tie there - it's meant to support CLI scenarios. Having a user run it in a separate console session is probably the way to go as the watcher keeps running until explicitly terminated (doesn't run in the background).

peterblazejewicz:
@sayedihashimi 
see why I've left bundler without change (at least for now):
https://github.com/OmniSharp/generator-aspnet/pull/768/commits/ac3145e4580b140269cf41b62cdce522731c0eee
We have to work out best UX for users with generator and Bundler (should we literally port templates solution based on VS experience?)

peterblazejewicz:
@mlorbetske Thanks!

PR #632, opened 2016-06-24
Remove DnxInvisibleContent Bower artifact references

scottaddie:
The "EmptyWeb" template's XPROJ file contains references to the following 2 Bower files:
1. `bower.json`
2. `.bowerrc`

Since neither of these files is included in the template, it's probably best to remove them from the `DnxInvisibleContent` item group altogether. I would expect a "bare bones" project which makes no assumptions with regard to a client-side package manager when I choose the empty project template.

### DIFF
--- a/src/BaseTemplates/EmptyWeb/EmptyWeb.xproj
+++ b/src/BaseTemplates/EmptyWeb/EmptyWeb.xproj
@@ -17,9 +17,5 @@
   <PropertyGroup>
     <SchemaVersion>2.0</SchemaVersion>
   </PropertyGroup>
-  <ItemGroup>
-    <DnxInvisibleContent Include="bower.json" />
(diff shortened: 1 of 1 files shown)


phenning:
We actually left these in here intentionally for the empty template because we have the item template which adds them and there is no tooling to add this back to the xproj if those item templates are used.

scottaddie:
@phenning Are there plans to add tooling for this? After all, that's the value Visual Studio adds. If I have to manually edit a project file for a common task such as this, then Visual Studio has failed me.

phenning:
I'm not aware of any current plans. However, future releases are still being planned. Let's hold off on this PR until the roadmap is complete. It may be that we won't need the InvisibleContent item group in the future.

PR #633, opened 2016-06-24
Fix incorrect sorting caused by changing Microsoft.EntityFrameworkCor…

phenning:
…e.Design to Microsoft.EntityFrameworkCore.Tools

@Eilon, @bricelam  FYI.


phenning:
@peterblazejewicz FYI

Issue #226, opened 2015-10-18
Consider adding .editorconfig for contributors


peterblazejewicz:
The code authors using VS studio most probably use built-in settings for source files, and even if VS does not have built-in support for EditorConfig:
http://editorconfig.org/
and even if VS 2015 seems to have broken support for plugin supporting `.editorconfig`:
https://github.com/editorconfig/editorconfig-visualstudio/issues/24
it would be nice if the VS built-in settings used in project are shared and available to contributors.
We could use that config file in all other editors to avoid unwanted commit changes and diffs.
Thanks!

rustd:
This is not a templates issue. what change do you expect?

peterblazejewicz:
The MS Team probably uses default settings in VS when formatting code. The could be nice for people using other editors to move that settings into commonly used EditorConfig file (I'm using such solution at work). That's like manifesto on how code should be formatted and some editors respect it by default (or via extension).
For example I have always to check if I'm not introducing side changes in commits just by adding or removing white space, etc.
Also related project like `generator-aspnet` could adopt that configuration to avoid unnecessary changes when updating content. The `generator-aspnet` uses `.editorconfig` - I use it for generator related content only thought.
Thanks!

CADbloke:
ReBracer is another extension you can use to push standardized settings.

aspnet-hello:
All issues in this repo are being closed because this repo is no longer in use. Please see the readme for more information: https://github.com/aspnet/Templates/blob/dev/README.md.

PR #639, opened 2016-06-28
Templates to enforce `using`

IEvangelist:
Since the `IWebHost` implements `IDisposable`, let's make the templates enforce best practices.

### DIFF
--- a/src/BaseTemplates/EmptyWeb/Program.cs
+++ b/src/BaseTemplates/EmptyWeb/Program.cs
@@ -11,14 +11,15 @@ namespace $safeprojectname$
     {
         public static void Main(string[] args)
         {
-            var host = new WebHostBuilder()
-                .UseKestrel()
-                .UseContentRoot(Directory.GetCurrentDirectory())$if$ ($aspnet_useplatformhandler$ == false)$else$
-                .UseIISIntegration()$endif$
-                .UseStartup<Startup>()
-                .Build();
-
-            host.Run();
+            using (var host = new WebHostBuilder()
+								 .UseKestrel()
+								 .UseContentRoot(Directory.GetCurrentDirectory())$if$ ($aspnet_useplatformhandler$ == false)$else$
+								 .UseIISIntegration()$endif$
+								 .UseStartup<Startup>()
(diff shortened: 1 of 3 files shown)


phenning:
@davidfowl thoughts?

peterblazejewicz:
2c: The best practise documentation (and sample code) usually are simple, short and explicit:
https://msdn.microsoft.com/en-us/library/3bwa4xa9(v=vs.110).aspx
e.g.:

``` csharp
using (StreamReader s = new StreamReader("File1.txt")) {
```

but not:

```
using(var a = new b()
  .UseFeatureC()
  .UseFeatureD()
  .UseAsFeatureE()
  .Build())
```

In such complex code there is no obvious hint which method returns IDisposable: contructor, extension methods, build method? It obviates possible gain from changing a code IMO.
The linked docs states that there are **two** ways to handle IDisposable BTW (depending on contex, preferences, etc)

@IEvangelist white space should not to be changed, see: https://github.com/aspnet/Templates/pull/639/files?w=1
@phenning https://github.com/aspnet/Templates/issues/226

IEvangelist:
@peterblazejewicz So would you prefer to have a `try / finally`? I guess my point is that I want the `IDisposable` to be honored. It belongs in either a `using` or in a `try / finally`.

peterblazejewicz:
No, I don't have preferences here (I use a lot of `using` at work though). Let's wait for @davidfowl 💬

davidfowl:
We debated this a bit and moved away from it. Run disposed the appropriate things.

IEvangelist:
@davidfowl I see why now. It certainly is **not** obvious from the templates -- you have to go looking at [the implementation of `Run`](https://github.com/aspnet/Hosting/blob/7b513de5c3a8379706a76c6b7bb0380ee83194b7/src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs), which is an extension method, which internally wraps its usage in a `using`.

davidfowl:
Yep! It's a bit cleaner to look at since wrapping that many lines of code in a using is pretty awful.

PR #641, opened 2016-06-29
Remove unused GetCurrentUserAsync method in AccountController

### LINKED ISSUE #634: GetCurrentUserAsync is not used in AccountController
https://github.com/aspnet/Templates/blob/dev/src/Rules/StarterWeb/IndividualAuth/Controllers/AccountController.cs#L449-L452

phenning:
Fixes #634

### DIFF
--- a/src/Rules/StarterWeb/IndividualAuth/Controllers/AccountController.cs
+++ b/src/Rules/StarterWeb/IndividualAuth/Controllers/AccountController.cs
@@ -446,11 +446,6 @@ namespace $safeprojectname$.Controllers
             }
         }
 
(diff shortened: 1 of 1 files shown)


phenning:
@joeloff

peterblazejewicz:
This removes only method defintion, but the method **is used** in templates:
https://github.com/aspnet/Templates/search?q=GetCurrentUserAsync&type=Code&utf8=%E2%9C%93
maybe the scope of PR is not correct (too narrow)?

phenning:
@peterblazejewicz The method in the account controller is not used by any method in the account controller. The manage controller has it's own method which is used by other methods in that controller. This actually has been unused in the account controller for a few releases.

Issue #660, opened 2016-08-11
Fix external login flow for failed/partial authentication


Eilon:
Original bug: https://github.com/aspnet/Identity/issues/915
# 

We want to make the change described in option (3) in @Tratcher 's comment at https://github.com/aspnet/Identity/issues/915#issuecomment-238939350.

Specifically, if the user attempts to use an OAuth login and they still have the external login cookie, the template code will automatically pick up where it left off and complete the flow.

Eilon:
We can have @troydai do the fix to the templates.

mikes-gh:
We need to account for Identity.External being a different provider.
@Tratcher solution didn't allow for this.
TODO : refactor out repeated code

```
// POST: /Account/ExternalLogin
[HttpPost]
[AllowAnonymous]
[ValidateAntiForgeryToken]
public async Task<IActionResult> ExternalLogin(string provider, string returnUrl = null)
{
    var info = await _signInManager.GetExternalLoginInfoAsync();

    // No External Cookie (normal flow)
    if (info == null)
    {
        var redirectUrl = Url.Action("ExternalLoginCallback", "Account", new { ReturnUrl = returnUrl });
        var properties = _signInManager.ConfigureExternalAuthenticationProperties(provider, redirectUrl);
        return new ChallengeResult(provider, properties);
    }
    else
    {

        // External Cookie present same provider
        if (info.LoginProvider == provider)
        {
            return RedirectToAction(nameof(AccountController.ExternalLoginCallback), "Account");
        }

        // External Cookie present different provider
        else
        {
            await _signInManager.SignOutAsync();
            return RedirectToAction(nameof(AccountController.ExternalLoginRedirect), "Account", new { provider = provider, returnUrl = returnUrl });
        }
    }   
}

[HttpGet]
[AllowAnonymous]
public IActionResult ExternalLoginRedirect(string provider, string returnUrl = null)
{
    var redirectUrl = Url.Action("ExternalLoginCallback", "Account", new { ReturnUrl = returnUrl });
    var properties = _signInManager.ConfigureExternalAuthenticationProperties(provider, redirectUrl);
    return new ChallengeResult(provider, properties);
}
```

troydai:
Thanks, @mikes-gh . A good point. I'll make sure different provider scenario is handled.

Tratcher:
RE: Challenging for provider A does not currently work if you're already logged in with provider B and they share the same cookie: https://github.com/aspnet/Security/issues/859

[...]

PR #662, opened 2016-08-16
Update AccountController for partial auth case

troydai:
Clear the external cookie in Login Action when it exists.

Issue: #660 

/cc @phenning @Tratcher @HaoK @Eilon

### DIFF
--- a/src/Rules/StarterWeb/IndividualAuth/Controllers/AccountController.cs
+++ b/src/Rules/StarterWeb/IndividualAuth/Controllers/AccountController.cs
@@ -8,6 +8,7 @@ using Microsoft.AspNetCore.Identity;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.Rendering;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 using $safeprojectname$.Models;
 using $safeprojectname$.Models.AccountViewModels;
 using $safeprojectname$.Services;
@@ -22,16 +23,19 @@ namespace $safeprojectname$.Controllers
         private readonly IEmailSender _emailSender;
         private readonly ISmsSender _smsSender;
         private readonly ILogger _logger;
+        private readonly string _externalCookieScheme;
 
         public AccountController(
             UserManager<ApplicationUser> userManager,
             SignInManager<ApplicationUser> signInManager,
+            IOptions<IdentityCookieOptions> identityCookieOptions,
             IEmailSender emailSender,
             ISmsSender smsSender,
             ILoggerFactory loggerFactory)
         {
             _userManager = userManager;
             _signInManager = signInManager;
+            _externalCookieScheme = identityCookieOptions.Value.ExternalCookieAuthenticationScheme;
             _emailSender = emailSender;
             _smsSender = smsSender;
             _logger = loggerFactory.CreateLogger<AccountController>();
@@ -43,6 +47,9 @@ namespace $safeprojectname$.Controllers
         [AllowAnonymous]
         public IActionResult Login(string returnUrl = null)
         {
+            // Clear the existing external cookie to ensure a clean login process
+            await HttpContext.Authentication.SignOutAsync(_externalCookieScheme);
+
             ViewData["ReturnUrl"] = returnUrl;
             return View();
         }


src/Rules/StarterWeb/IndividualAuth/Controllers/AccountController.cs
```diff
@@ -43,6 +47,12 @@ public class AccountController : Controller
         [AllowAnonymous]
         public IActionResult Login(string returnUrl = null)
         {
+            // Clear the existing external cookie to ensure a clean login process
+            if (await HttpContext.Authentication.AuthenticateAsync(_externalCookieScheme) != null)
```
HaoK: Might be simpler to just clear the cookie instead of checking?   Authenticate might not be cheap either?

troydai: Tested and it works, too. I'll remove the test here.

phenning:
Looks good to me. :shipit:

Eilon:
Assuming the code is correct from a _behavior_ perspective (which looks fine to me, per our discussions), the code looks good to me from a _template_ perspective, so :shipit:

HaoK:
Its too bad the IdentityOptions are internal on the manager, would have been nice to avoid pulling in the IOptions, but oh well :shipit:

Tratcher:
:shipit:

gdoron:
I might be missing something here, but is the fix is just adding 

```
await HttpContext.Authentication.SignOutAsync(_externalCookieScheme);
```

to `Login`?

If so, how would that help applications where users don't need to go a designated login page.
We for example, have a SignIn button which opens a modal window with email-password or external oauth providers.

And what about SPAs? They don't have a Login page at all.

Tratcher:
@gdoron those are different flows that may or may not be impacted by similar issues.  e.g. isn't your modal window just the login page in a pop-up window?  As opposed to SPA apps that don't even use the two cookie approach, so I don't see how they'd be affected by this issue.

gdoron:
@Tratcher 

> isn't your modal window just the login page in a pop-up window

Unfortunately, no. Because of strict  timeline and outsource of the client side development, we have that modal window all the time on the page but hidden (it's embedded as a partial view into Layout.cshtml)
So the fix for me is issue an AJAX request to some action that just have this:

```
await HttpContext.Authentication.SignOutAsync(_externalCookieScheme);
```

when the modal window is opened? What exactly does it do?

Also I didn't understand your comment regarding SPA, why wouldn't a SPA use the auth cookies?

Thanks.

Tratcher:
That line deletes the temporary external-provider cookie if it exists.

@phenning do we have a SPA template with individual auth to demonstrate the flow? Or is that what @javiercn is building an STS for?

Eilon:
There is no SPA template. Even the new one with an STS won't _be_ a SPA, though we are ensuring that it can be used in SPA scenarios (you have to write the SPA yourself, though, and call the STS).

mikes-gh:
I thought we decided anything in login action is not guaranteed to run?
The back button for example by passes this fix.

troydai:
@mikes-gh If the end user choose to sign in through external identity provider, the login page will then redirect to external identity provider and then redirect back to the `ExternalLoginCallback`. At this point if anything goes wrong, it's really unnatural to click back button to navigate to the Login page. Because you need to keep clicking a exact number of times in a fast pace to avoid redirect back. (I tried. For application create from template you need to click twice.)

The other scenarios have a larger issue that it requires an additional GET action to be invoked to successfully clear the external cookie in the case that end user decided to switch external providers. Clearing the external cookie head on in the beginning of the log in process is the cleanest solution for template.

mikes-gh:
Unfortunately this does not fix the issue.
I can't disable my users back button.
I do feel this issue had plenty of feedback that wasn't used.
In the end we have a change that several users said wont work in real world backed by sound examples and reasoning.
It's simply too easy to get to a cached version of the login page and hence bypass the action.
It's very easy to get a patial login and in my testing with real users 

You only need to click back once by the way to repo.

gdoron:
Agreed, it's really not the fix I was hoping for.
An extra GET request isn't as bad as a fix that might not work and lock our users out.

Issue #643, opened 2016-07-01
[Identity] ResetPassword is using FindByNameAsync, not FindByEmailAsync


brockallen:
Moved here from: https://github.com/aspnet/Identity/issues/865

The template is using the wrong API to find the account for password resets. It should be based on email, not the username. This makes it easier for people who use the template and don't use email for the username.

phenning:
Fixed by PR #665

brockallen:
Thanks.

PR #665, opened 2016-08-19
IndividualAuth password resets: get user by email

501st-alpha1:
Since the user is prompted for an email to reset their password, use
that to look up their user account, rather than the `UserName`
field. This makes a difference if you have decoupled those two fields,
allowing the user to specify their username rather than storing the
email to both.


phenning:
Thank you for the contribution!

phenning:
Closes #643 643

PR #668, opened 2016-08-31
Update MVC, Kestrel and EF versions to 1.0.1. Explictly add Routing 1…

phenning:
….0.1 reference

//cc @muratg @rynowak @joeloff @divega @mlorbetske 

Attached is the closure project.json which will show the versions that the templates are using.

[project.json.txt](https://github.com/aspnet/Templates/files/448439/project.json.txt)


phenning:
Ignore the massive number of frameworks targeted in the attached project.json, this json is just used to make sure we have a complete closure with the cache msi. we don't ship this json file.

divega:
EF package version changes look good assuming AspNetCorePackageVersion is 1.0.1.

rynowak:
Changes for Kestrel, Routing and MVC look right.

For anyone other than @phenning that's curious, we don't need to add antiforgery to the templates. MVC 1.0.1 references 1.0.1 of antiforgery, because if you want the FIPS fix you have to move both to 1.0.1.

phenning:
Merging update.

Issue #563, opened 2016-05-12
Remove unused `Microsoft.IdentityModel.Protocols.OpenIdConnect` using statement from Work and School templates


dougbu:
Bad enough that most of the C# files in this template have unnecessary `using`s but why does the `Startup` class use `Microsoft.IdentityModel.Protocols.OpenIdConnect`? The `UseOpenIdConnectAuthentication()` extension isn't in that namespace and that looks like all that's used here.

Of course, close this as a dupe if a general `Remove and Sort Usings` is already planned.

phenning:
We should remove the Microsoft.IdentityModel.Protocols.OpenIdConnect using statement.

Do you see any other "unused" namespaces other then the following? We include these in all template files as a matter of course in all CS files.

```
using System;
using System.Collections.Generic;
using System.Linq;
using System.Threading.Tasks;

```

dougbu:
> We include these in all template files as a matter of course in all CS files.

Why is this "a matter of course"? Unused means unnecessary here.

The first thing VS does if you open one of these files is offer to remove unused `using`s. Right-thinking editors like VS Code go further and include unused `using`s in the warnings it displays. And IntelliSense will add anything that's needed right back -- if that ever happens.

![usings](https://cloud.githubusercontent.com/assets/6431421/15203272/cd9fc8d2-17b6-11e6-9f58-e0e6ad3ad66c.png)

phenning:
My understanding is that the reason we want those particular namespaces in there for discoverability and as a convenience for those common namespaces which customers would most likely use when expanding the template content. We actually have an explicit build step to enforce that these namespaces are present and in order.

In any event, removing anything other then the OpenIdConnect namespace would likely not be something we would take for preview1

//cc @Eilon, @DamianEdwards @rustd

phenning:
TFS [220979](https://devdiv.visualstudio.com/defaultcollection/DevDiv/_workitems?id=220979&_a=edit)

PR #672, opened 2016-09-15
Remove unused Microsoft.IdentityModel.Protocols.OpenIdConnect using s…

phenning:
…tatements

Still needed for the AD-Read versions of the templates since we use OpenIdConnectResponseType in those templates.

@mlorbetske @joeloff


phenning:
This Fixes #563

PR #673, opened 2016-09-27
Recommend HTTPS by a commented out attribute

vanillajonathan:
Using HTTPS is important for security.

We cannot set this attribute by default since the Visual Studio local web development environment is not configured to use HTTPS. But we can add a commented out attribute to encourage the developer to enable HTTPS for the Account and Manage controllers.


Eilon:
Closing this because we plan to add HTTPS to all the templates anyway. HTTPS is good 😄

PR #677, opened 2016-10-12
Fixes issue when creating full framework projects, we use netcoreapp1…

phenning:
….0 instead of net4xx

@joeloff @mlorbetske


joeloff:
:shipit:

PR #699, opened 2016-11-04
Add bundler minifier and EF tools

joeloff:


joeloff:
@balachir @barrytang @seancpeters

src/BaseTemplates/StarterWeb/StarterWeb.csproj
```diff
@@ -69,33 +69,11 @@
...
-      "appsettings.json",
-      "web.config"
-    ]
-  },
 
-  "scripts": {
-    "prepublish": [ "bower install", "dotnet bundle" ]
-  }
-  -->
+  <Target Name="PrepublishScript" BeforeTargets="Publish">
```
barrytang, line 80: This needs to be below the following Import.

barrytang (changes requested):

src/BaseTemplates/StarterWeb/StarterWeb.csproj
joeloff: Fixed in the next commit

PR #701, opened 2016-11-04
Switch templates to use .AddUserSecrets(string)

natemcmaster:
Build passes. Not sure how to test if I've used the right variable replacement. `"aspnet-$safeprojectname$-$guid2$"`

cc @barrytang @joeloff


joeloff:
@natemcmaster You should be able to edit the file under Common7\IDE\WebTemplates where the files are dropped and create a new project to see if they are replaced.

natemcmaster:
Using dev15? I'm on macos using VS Code :-/

joeloff:
The replacement parameters are valid, so this should be fine.

:shipit:

PR #702, opened 2016-11-04
Make EmptyWeb vstemplate consistent with StarterWeb

joeloff:
Removing OpenInEditor as that causes a hang


joeloff:
@balachir @barrytang

balachir:
:shipit:

PR #707, opened 2016-11-11
Sepeters 286353 line endings

seancpeters:
Fixed inconsistent line endings in starterweb.csproj


seancpeters:
This fix was chasing a red herring, the real problem seems to be in VS template processing.

PR #712, opened 2016-11-16
Sepeters template settings coherence path

seancpeters:
Part of the fix for getting updatepkgs.ps1 to correctly copy everything.


seancpeters:
PR had some cruft in it. deleting and re-creating

PR #721, opened 2016-11-22
Remove BundlerMinifier from template

phenning:
@mlorbetske, @joeloff


joeloff:
Does this mean we  need to remove it from the feed installer too?

phenning:
Yes, we should remove it from the feed installer after this PR goes in.

joeloff:
:shipit:

PR #729, opened 2016-11-23
Add reference to EFCore.Design package in Individual Auth

prafullbhosale:
cc @mlorbetske @phenning 

This reference is needed for using EF tools from the package manager console.


mlorbetske:
:shipit:

PR #731, opened 2016-11-30
Update globbing patterns and use SDK attribute

mlorbetske:
@phenning @seancpeters @davidfowl

### DIFF
--- a/src/BaseTemplates/EmptyWeb/EmptyWeb.csproj
+++ b/src/BaseTemplates/EmptyWeb/EmptyWeb.csproj
@@ -7,14 +7,24 @@
     <RuntimeIdentifier>win7-x86</RuntimeIdentifier>$endif$
   </PropertyGroup>
 
+  <PropertyGroup>  
+    <GlobalExclude>$(GlobalExclude);bin\**;obj\**;node_modules\**;jspm_packages\**;bower_components\**;**\*.user;**\*.*proj</GlobalExclude>
+  </PropertyGroup>
+
   <ItemGroup>
-    <Compile Include="**\*.cs" Exclude="@(GlobalExclude)" />
-    <EmbeddedResource Include="**\*.resx" Exclude="@(GlobalExclude)" />
-    <Content Include="@(ContentFiles)" Exclude="@(GlobalExclude);@(ContentExclude);@(MiscFiles)" />
-    <Content Include="@(PublishFiles)" Exclude="@(GlobalExclude);@(ContentExclude);@(MiscFiles)" CopyToPublishDirectory="PreserveNewest" />
-    <None Include="@(MiscFiles)" Exclude="@(GlobalExclude);@(ContentExclude);@(ContentFiles)" CopyToPublishDirectory="PreserveNewest" />
+    <None Include="**\*" />
+    <Compile Include="**\*.cs" />
+    <EmbeddedResource Include="**\*.resx" />
     <Content Include="wwwroot\**" CopyToPublishDirectory="PreserveNewest" />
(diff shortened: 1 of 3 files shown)


davidfowl:
Is the SDK thing just in prep for that change? Because it's incomplete, the props and targets are supposed to be gone as a result of the SDK node being there. Also the fact that the web templates globbing patterns got crazy makes me sad.

mlorbetske:
I'll remove the extra props & targets, sending you mail on what we're thinking post this upcoming release

src/BaseTemplates/EmptyWeb/EmptyWeb.csproj
```diff
@@ -9,8 +8,12 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <Compile Include="**\*.cs" />
-    <EmbeddedResource Include="**\*.resx" />
+    <Compile Include="**\*.cs" Exclude="$(GlobalExclude)" />
```
phenning: Just checking, the re-addition of all the Excludes of GlobalExclude is intentional, correct?

phenning:
It looks like the removal of the Sdk.Web PackageReference isn't there for StarterWeb.csproj. Both WebAPI and Empty do show this as removed.

phenning:
Looks good :shipit:

phenning:
Other then this being really ugly, assuming it addresses the perf concerns, looks good from a template perspective.

:shipit:

mlorbetske:
@phenning ugly in a way we can change? Yes, that's to fix the perf issues

src/BaseTemplates/EmptyWeb/EmptyWeb.csproj
```diff
@@ -8,10 +7,24 @@
...
+    <None Include="**\*" />
     <Compile Include="**\*.cs" />
     <EmbeddedResource Include="**\*.resx" />
+    <Content Include="wwwroot\**" CopyToPublishDirectory="PreserveNewest" />
+    <Content Include="**\*.cshtml" Exclude="wwwroot\**\*.cshtml" CopyToPublishDirectory="PreserveNewest" />
+    <Content Include="**\*.config" Exclude="wwwroot\**\*.config" CopyToPublishDirectory="PreserveNewest" />
+    <Content Include="**\*.json" Exclude="wwwroot\**\*.json" CopyToPublishDirectory="PreserveNewest" />
     <Folder Include="wwwroot\" />
+
+    <None Remove="@(Content);@(Compile);@(EmbeddedResource);$(GlobalExclude)" />
```
davidfowl: Does this part have to be in the project file too?

mlorbetske: @davidfowl we can move the removes to the targets

mlorbetske: https://github.com/aspnet/websdk/pull/69

phenning: @mlorbetske I don't think so, at least for now.

mlorbetske: @davidfowl the only issue with moving these to the targets is that CPS will no longer know where to put changes that override them, but this can be worked around if the user changes their csproj to remove the item from the item group & put it in to the right one before the targets run

phenning (commented):
This looks better without the removes.

mlorbetske:
Merging this to unblock broader testing, will submit another PR if needed

davidfowl:
@mlorbetske Do you know why the Reference items are needed for .NET Framework?

PR #740, opened 2016-12-09
Update to 1.0.3 ASP.NET Core Runtime

phenning:
* Update to 1.0.3 ASP.NET Core Runtime
* Add TemplateClosure.csproj to validate Local Feed msi
* Removed the ToolsFile resolution in settings since we just use the scaffolding sem ver for all tool versions

@joeloff @mlorbetske

### DIFF
--- a/Templates.Settings.targets
+++ b/Templates.Settings.targets
@@ -60,19 +60,14 @@
     
     <Target Name="DeterminePackageVersions" DependsOnTargets="SetupPackageSource">
         <ItemGroup Condition="'$(CoherencePackageSources)' != ''">
-            <CoreAppFile Include="$(PackageSourceExternal)\Microsoft.NETCore.App.*.nupkg"/>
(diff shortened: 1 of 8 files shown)


phenning:
verified that package closure is good with the current local cache build.

Templates.Settings.targets
```diff
@@ -60,20 +60,15 @@
     
     <Target Name="DeterminePackageVersions" DependsOnTargets="SetupPackageSource">
         <ItemGroup Condition="'$(CoherencePackageSources)' != ''">
-            <CoreAppFile Include="$(PackageSourceExternal)\Microsoft.NETCore.App.*.nupkg"/>
```
joeloff, line 63: These aren't needed anymore?

phenning: Nope, since we aren't picking up semver'd versions of this, but instead the static version, we just set the version to 1.0.3 directly.. see line 99

joeloff:
:shipit:

PR #744, opened 2016-12-13
Make the comment in web.config multiple lines

### LINKED ISSUE #718: Make the comment in web.config multiple lines
In the MVP summit we did a UX study and found that people don't read the comment in web.config about not adding appSettings. If it was more prominent it would be more likely to be read. Also @Andrew-MSFT suggested actually adding <appSettings> in the comment block.

How about this.

```xml
<!--

Configure your application settings in appsettings.json. Learn more at https://go.microsoft.com/fwlink/?LinkId=786380

  <appSettings>
      appSettings in web.config is NO longer being used in ASP.NET Core, instead use appsettings.json
  </appSettings>
-->
```

phenning:
Fixes #718 

@sayedihashimi @mlorbetske

### DIFF
--- a/src/BaseTemplates/EmptyWeb/web.config
+++ b/src/BaseTemplates/EmptyWeb/web.config
@@ -2,7 +2,11 @@
 <configuration>
 
   <!--
-    Configure your application settings in appsettings.json. Learn more at https://go.microsoft.com/fwlink/?LinkId=786380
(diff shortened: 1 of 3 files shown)


mlorbetske (approved):

sayedihashimi:
Thanks Phil

PR #764, opened 2016-12-19
Phenning/ai cleanup

phenning:
@mlorbetske, @pakrym 

Got rid of the AIPatchPackageVersion, instead just use the existing property

Remove other no longer needed files.
Update the Shared _Layout.cshtml files


phenning:
Merging this now to fix issue with incorrect stamped package version in current builds

PR #765, opened 2016-12-19
Add appsettings.Development.json to override certain logging settings

### LINKED ISSUE #708: Only add the ConsoleLogger if IHostingEnvironment.IsProduction() is false
We should change all the templates so that the console logger is only added in `Startup.Configure` **if** the current environment is not production, e.g.:
``` c#
public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerFactory)
{
    if (!env.IsProduction())
    {
        loggerFactory.AddConsole(Configuration.GetSection("Logging"));
        loggerFactory.AddDebug();
    }
    ...
}
```

phenning:
Closes #708 

@DamianEdwards @davidfowl @mlorbetske


DamianEdwards:
@divega 
Looks good to me

DamianEdwards (approved):
Looks good to me

PR #775, opened 2016-12-30
Fix trailing whitespace

phenning:
@mlorbetske @seancpeters 

Gets the CI build off the floor.


phenning:
Optimistically merging.

PR #787, opened 2017-01-11
Re-add netcore.app to templates for WebCore projects.

phenning:
Add NETCore.App back to template for WebCore projects


phenning:
@mlorbetske @seancpeters

PR #792, opened 2017-01-17
Remove references to Microsoft.AspNetCore.Mvc.Razor.ViewCompilation f…

pranavkm:
…rom templates


phenning (approved):
We will need to remove from the CLI templates as well.

pranavkm:
@phenning we verified that the CLI templates do not include this package.

PR #798, opened 2017-01-22
[Fixes #797] Preserve redirectUrl in ExternalLoginCallback error

### LINKED ISSUE #797: ReturnUrl is lost if GetExternalLoginInfoAsync fails in AccountController.ExternalLoginCallback
It's probably not very likely that it will occur but as it attempts to recover by sending you back to the login page rather than erroring out it should preserve the return url.

rtpHarry:
[Fixes #797]


aspnet-hello:
All issues in this repo are being closed because this repo is no longer in use. Please see the readme for more information: https://github.com/aspnet/Templates/blob/dev/README.md.

PR #799, opened 2017-01-23
Update to 1.1.1 MVC package

phenning:
@mlorbetske @Eilon


phenning:
Will push down into dev after merging

Eilon (approved):
The change looks good, but we don't know yet if we're going to take this change right now. Will keep everyone up-to-date on that.

mlorbetske (approved):

PR #805, opened 2017-01-26
Disable Browserlink for 1.1.0 projects.

phenning:
@jodavis @joeloff @mlorbetske

### DIFF
--- a/src/BaseTemplates/StarterWeb/Startup.cs
+++ b/src/BaseTemplates/StarterWeb/Startup.cs
@@ -40,10 +40,9 @@ namespace $safeprojectname$
             if (env.IsDevelopment())
             {
                 app.UseDeveloperExceptionPage();$if$ ($platformversion$ == 1.1)
-                // To enable Browser Link, uncomment the line below and add the
-                // Microsoft.VisualStudio.Web.BrowserLink package.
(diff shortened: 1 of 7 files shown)


phenning:
@seancpeters

src/BaseTemplates/StarterWeb/Startup.cs
```diff
@@ -39,8 +39,12 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerF
 
             if (env.IsDevelopment())
             {
-                app.UseDeveloperExceptionPage();
-                app.UseBrowserLink();
+                app.UseDeveloperExceptionPage();$if$ ($platformversion$ == 1.1)
```
joeloff, line 42: Should this be >= 1.1. IF the version is 1.1.1 then the condition would be false and it will still enable BrowserLink, no?

phenning: This corresponds with the versions declared here:

https://github.com/aspnet/Templates/blob/dev/src/Templates.xml#L401-L402

joeloff: Chatted with Phil, this makes more sense now

joeloff:
:shipit:

PR #810, opened 2017-02-01
Remove the ApplicationInsights.AspNetCore package redirection for 1.1

phenning:
The redirect was wrong, plus we never actually needed it as there is no 1.1 specific version of this package

@mlorbetske @joeloff


mlorbetske (approved):

joeloff:
:shipit:

PR #811, opened 2017-02-02
Add MVC precompilation package for 1.1

mlorbetske:


src/BaseTemplates/StarterWeb/StarterWeb.csproj
```diff
@@ -2,7 +2,8 @@
 
   <PropertyGroup>
     <TargetFramework>$netframeworkversion$</TargetFramework>$if$ ($context$ == WebCoreFullFramework)
-    <RuntimeIdentifier>win7-x86</RuntimeIdentifier>$endif$
+    <RuntimeIdentifier>win7-x86</RuntimeIdentifier>$endif$$if$ ($netframeworkversion$ == netcoreapp1.1)
```
phenning: This needs to be $platformversion$, because $netframeworkversion$  will be set to net452, etc for Core on Full Fx

src/BaseTemplates/StarterWeb/StarterWeb.csproj
```diff
@@ -2,7 +2,8 @@
...
+    <MvcRazorCompileOnPublish>true</MvcRazorCompileOnPublish>$endif$
```
phenning: Is auto enabling of the feature no longer done? Or are we just putting this here for discoverability?

phenning (changes requested):

src/BaseTemplates/StarterWeb/StarterWeb.csproj
seancpeters: My understanding is that for MVC templates using netcoreapp1.1 the razor functionality will always be enabled.

phenning (approved):

seancpeters:
:shipit:

phenning:
Approved.. but we need to hold off the merge until we update feed installer to carry the package.

PR #812, opened 2017-02-02
This adds creation of a few "Closure" projects which we can use to te…

phenning:
…st the local feed.

@mlorbetske 

Verified generated template.xml is identical to previous to this change, added closure projects to verify local feed against.

Cleaned up some naming.


phenning:
There are additional scripts to restore these projects which will be checked into WTE.

phenning:
Merging this to shake out any CI issues.

PR #822, opened 2017-02-17
Fix incorrect versions in 1.1 Closure projects

phenning:
@mlorbetske


phenning:
Merging this now as it is not dependent on any of the other pending changes

PR #844, opened 2017-04-25
Add validate scopes to templates

pakrym:
https://github.com/aspnet/DependencyInjection/issues/481
/cc @DamianEdwards


phenning:
Thus should probably be applied to dotnet/templating.

PR #862, opened 2017-08-21
Use template engine templates

phenning:


phenning:
initial merge to dev, more cleanup and CI work is needed

